PR #14, opened 2021-01-06
Minor documentation edits

nickjalbert:
A few small edits:
* `agent.step()` -> `agent.advance()`
* Capitalize Python (believe this is standard)
* Remove trailing whitespace
* "Codify" a few things that seemed reasonable to codify (e.g. `False` when referring to the Python boolean)
* A few minor rewrites for clarity

How would you feel if I reflowed the paragraphs to be 80 chars max per line?  Makes editing and diffs easier to read IMO.

### DIFF
--- a/documentation/architecture_and_design.rst
+++ b/documentation/architecture_and_design.rst
@@ -5,16 +5,16 @@ Architecture and Design
 Background from Reinforcement Learning
 ======================================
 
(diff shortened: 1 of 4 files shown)


andyk:
Thanks! Merged these changed. Also, reflowing the paragraphs to be 80 chars max per line would good!

PR #16, opened 2021-01-10
Fix documentation search

nickjalbert:
Weirdly enough, the problem is still the same as discussed here: https://github.com/sphinx-doc/sphinx/issues/5460

This is a quick and dirty fix that copies the Sphinx RTD theme layout and adds the missing import.  This is bad because when updating the RTD theme layout package, you have to re-copy the layout and re-add the import (at least until they fix the bug).

There's probably a more maintainable way to add the missing import;  I'll do a little poking around Sphinx and the RTD theme to figure it out.


nickjalbert:
closed because #17 is a better way to fix search

PR #20, opened 2021-01-11
Update requirements for building docs and add helper scripts

nickjalbert:
When I merged in upstream, I realized I could no longer build the docs.  Turns out I needed to:

* Add the `agentos` module to my `PYTHONPATH`, and
* Install `gym` and `mlflow`

Presumably sphinx is trying to do something like import these modules (and their dependencies) to get at their docstrings.

Not sure how you want to handle this, but adding those two packages to the `documentation/requirements.txt` is one way!

Also, added some convenience scripts for building and cleaning:

* `./documentation/scripts/build-docs`
* `./documentation/scripts/clean-docs`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,17 +51,19 @@ To run tests::
 Building Documentation / Agentos.org Website
 --------------------------------------------
 
-The documentation is in the ``docs`` direcory and written in `ReStructuredText <https://docutils.sourceforge.io/rst.html>`_.
-To build the docs you'll need to use `Sphinx <https://www.sphinx-doc.org>`_:::
+The documentation is in the ``docs`` directory and written in
+`ReStructuredText <https://docutils.sourceforge.io/rst.html>`_.
+To build the docs you'll need to use
+`Sphinx <https://www.sphinx-doc.org>`_:::
 
   pip install -r documentation/requirements.txt
-  sphinx-build docs docs/_build  # also checkout the pip package sphinx-autobuild
+  ./documentation/scripts/build-docs
   # Open and inspect docs/_build/index.html in your broswer.
 
 `agentos.org <https://agentos.org>`_ is a github.io website where the AgentOS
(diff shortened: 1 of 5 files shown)


andyk:
I've been thinking about trying to keep things as windows friendly as possible.

What do you think of using python scripts instead of bash scripts to keep things more portable?

I had added documentation/build_docs.py as a starting spot for a docs build script. I didn't document it (or test it much) yet though.

It uses the version file to put docs in a subdir w/ version as its name (which we can do w/ python too), which is how I was thinking we could eventually handle keeping past versions of docs available (and then make agentos.org redirect to agentos.org/latest or something).

And I'm not sure if Popen just works on windows. it also doesn't currently solve the problem of needing agentos to be on on the PYTHONPATH for the Sphinx autodocs feature to work.

andyk:
Also, I think Sphinx has a recommended way of handling the path problem via conf.py. See https://github.com/agentos-project/agentos/blob/39e2942ca2ebd080a1ef57ee9134fe061325a29a/documentation/conf.py#L9

I just was sloppy and didn't grok how it is supposed to work when I uncommented out line 17 so that I could make sure it's adding the necessary modules. though I did add that print statement with the typo on line 16 and forgot to delete it, and it showed the agentos dir on my pythonpath (maybe because I had used `pip install -e` at some point or some other reason).

In any case maybe we could fix line 17 (and delete line 16, the print statement) to solve the python path issue.

Re `mlflow` and `gym` in docs requirements, instead of adding them directly, could we require `-e ..` to have them get included transitively (then we don't need to maintain agentos's dependencies in two places)? This is what we do in the example agents currently, e.g., see https://github.com/agentos-project/agentos/blob/master/example_agents/evolutionary_agent/requirements.txt

andyk:
I think these issues are resolved after the commits today.

nickjalbert:
Yep!  Works for me now!

PR #32, opened 2021-01-14
Documentation updates

nickjalbert:
This PR contains the following:

* Adds the documentation build directory to gitignore (let me know if this breaks the doc publishing flow and I'll revert)
* Fix a `build_docs.py` crash if the `latest` symlink doesn't exist
* Update `build_docs.py` docstring
* Require Python 3.5 - 3.8 because 3.9 breaks the tensorflow requirement
* Capitalize Python in the text


nickjalbert:
Ok @andyk, updated `build_docs.py` so that it doesn't automatically install requirements.

.gitignore
```diff
@@ -9,3 +9,4 @@ build
 dist
 documentation/_build
 website/_site
+docs
```
andyk, line 12: This is a decent solution. If we commit this, then after the next time we merge the `master` branch into the `website` branch, we can just remove this entry in a commit only on the `website` branch. Otherwise it will be a bit more painful to push agentos.org website changes to production in the `website` branch.

Alternative A: what do you think about adding a `--release` flag to `build_docs.py` that causes the output to go to `docs/<curr_version>` and the `latest` symlink to be updated. And then, when the flag isn't passed have the standard output location be `documentation/_build` which is already ignored by git, and which I think is, by convention, the standard location of sphinx output?

Or Alternative B: we could have output go to `_build` by default and add a `--release_dir` flag to `build_docs.py` which the release master manually sets to build into the appropriate output dir and sets the `latest` symlink to point to it.

andyk (approved):
Left some comments about smallish things. Other than those, LGTM.

.gitignore
nickjalbert: I like Alternative A but I'll go with fixing up the website branch for expediency right now.  Maybe we could track Alternative A as a feature request?  Did we decide how to best track things?  I can create an issue if that would be useful.

andyk: Sounds good. 

Yeah I’ve been using GitHub issues, which integrates nicely with GitHub Projects, where I’ve been using a Kannan board to prioritize and track progress/status  of them.

nickjalbert: Filed an issue: https://github.com/agentos-project/agentos/issues/34

PR #33, opened 2021-01-14
use proper version format for alpha releases, document release process

andyk:


nickjalbert:
LGTM.  I'll merge this now and then fix the conflicts in my PR.

PR #35, opened 2021-01-15
Merges agentos/master into agentos/website and removes docs/ from gitignore

nickjalbert:
Let me know if there's a different way you think I should do this (or if we should skip PRs when merging master into website).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@ build
 dist
 documentation/_build
 website/_site
-docs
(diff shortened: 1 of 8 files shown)


andyk:
I think using PRs is a good sanity check

andyk:
Should we use separate PRs for the merge and pushing the updated built docs to docs/<version>?

nickjalbert:
I was thinking that it'd be cool to get the GitHub actions to automatically build the website and build the README

PR #36, opened 2021-01-15
Use os.sep and os.pardir

nickjalbert:
Minor updates to build_docs.py and build_readme.py that shouldn't change functionality:
* Use `os.sep` instead of `/`
* Use `os.pardir` instead of `..`
* Use `os.path.join` instead of format strings

Theoretically, should make these scripts more likely to run on windows (though, I haven't actually tested on windows)

### DIFF
--- a/documentation/build_docs.py
+++ b/documentation/build_docs.py
@@ -11,15 +11,22 @@ import os
 from subprocess import Popen
 
 docs_dir = os.path.dirname(os.path.abspath(__file__))
+version_file = os.path.join(docs_dir, os.pardir, 'agentos', 'version.py')
+loaded = SourceFileLoader('agentos.version', version_file).load_module()
+version = loaded.VERSION
 
-version = SourceFileLoader(
-    'agentos.version', os.path.join(docs_dir, '..', 'agentos', 'version.py')).load_module().VERSION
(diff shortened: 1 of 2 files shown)


andyk:
Good idea!

Is `os.pardir` relative to the file it’s called in or the path where the script is run from? I it’s the latter, then this wouldn’t run from outside the documentation dir any more. Did you test that by any chance? I had a similar issue before which is why I had added the `docs_dir` variable.

I created issue #47 — create a test that attempts to build the docs.

documentation/build_docs.py
```diff
@@ -11,15 +11,22 @@
 from subprocess import Popen
 
 docs_dir = os.path.dirname(os.path.abspath(__file__))
+version_file = os.path.join(docs_dir, os.pardir, 'agentos', 'version.py')
```
andyk, line 14: Good idea!

Is `os.pardir` relative to the file it’s used in or the path where the script is run from? If it’s the latter, then this wouldn’t run from outside the documentation dir any more. Did you test that by any chance? I had a similar issue before which is why I had added the `docs_dir` variable.

I created issue #47 — create a test that attempts to build the docs.

andyk: Duh, nevermind this is exactly what you’re doing!

andyk (approved):
LGTM

PR #40, opened 2021-01-15
Add scripts to format and lint code

nickjalbert:
Adds two Python scripts to automatically format and lint Python code.  These will specifically only run on Python files tracked by git (and not, for example, Python files in packages in a virtualenv you keep untracked in the git repo).

To format with black: `python scripts/format_code.py`

To lint with flake8: `python scripts/lint_code.py`

I specifically did **not**:
* Set up any particular rules for formatting and linting.  Let's figure out a reasonable configuration once this gets in (or we could just use defaults, which I think conform to the pep8 style guide)
* Run the formatter on the existing code.  Let's wait until we decide on how to configure the formatter.

Per our conversation earlier, the [black repo](https://github.com/psf/black) has 19k stars and 1k forks which makes me think it's reasonably popular in the community.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -11,6 +11,7 @@ sphinx_rtd_theme==0.5.1
 click>=7.0 # when updating, also update in ../setup.py
 
 # Tests requirements
+black==20.8b1
 cloudpickle==1.3.0  # gym 0.17.1 in setup.py requires cloudpickle<1.4.0,>=1.2.0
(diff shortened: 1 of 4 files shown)


andyk:
LGTM.

Wondering why the git check uses `Popen()` but the other subprocesses are created via `run()`. Maybe they way the it's using the return value or blocking?

andyk (approved):
LGTM

PR #43, opened 2021-01-17
Format and lint

nickjalbert:
This PR runs the black formatter over the repo and fixes any lint complaints (mostly unused imports and variables).  My (aesthetic) preference is a max line length of 79 like the standard library, but I'm not dead set on it if you have other preferences.

Lots of file changes, but mostly just tweaking the spacing of statements so this PR didn't take long to put together.  If you hate it, we can scrap it and go back to the drawing board and not much will be lost. 😄


nickjalbert:
I also tweaked the pytest configuration to spit out the run time of the tests which might be interesting for debugging purposes.  On my local box, test `test_evolutionary_agent` takes 90% of the time (129 sec of the 144 sec total run time).

Confirmed the tests pass and that the docs and readme build locally.

agentos/__init__.py
```diff
@@ -1,7 +1,5 @@
 """The ``agentos`` module provides an API for building learning agents."""
 
-from agentos.version import VERSION as __version__
```
nickjalbert, line 3: flake8 was complaining about this unused import, but is this something we need to make the package work right?

nickjalbert:
Tests are passing.  Merging per our conversation.

PR #47, opened 2021-01-23
Separate format and lint github actions

nickjalbert:
Breaks out the format and lint github actions into their own flows so the tests will still run even if black or flake8 raises an issue.


andyk:
Great thinking. LGTM (once tests finish)

andyk (approved):

Issue #48, opened 2021-01-23
Migrate build_docs.py and build_readme.py to the scripts dir


andyk:
They make more sense there. Developers will be more likely to notice and look for them there.

PR #52, opened 2021-01-23
Move documentation build scripts and fix link to github tests

nickjalbert:
* Found the link on the github testing badge was broken so fixed that in b1ff04a

agentosbadge

* Decided to knock out https://github.com/agentos-project/agentos/issues/48 while I was in there (although, I'll hold off on fixing "good first issues" in the future)

### DIFF
--- a/scripts/build_docs.py
+++ b/scripts/build_docs.py
@@ -3,20 +3,24 @@ Build the AgentOS documentation.
 
 To use::
 
-  $ python documentation/build_docs.py
+  $ python scripts/build_docs.py
 """
 
 from importlib.machinery import SourceFileLoader
 import os
 from subprocess import Popen
 
-docs_dir = os.path.dirname(os.path.abspath(__file__))
-version_file = os.path.join(docs_dir, os.pardir, "agentos", "version.py")
+scripts_dir = os.path.dirname(os.path.abspath(__file__))
+version_file = os.path.join(scripts_dir, os.pardir, "agentos", "version.py")
 loaded = SourceFileLoader("agentos.version", version_file).load_module()
 version = loaded.VERSION
 
-build_dir = os.path.normpath(os.path.join(docs_dir, os.pardir, "docs"))
+build_dir = os.path.normpath(os.path.join(scripts_dir, os.pardir, "docs"))
 versioned_build_dir = os.path.join(build_dir, f"{version}")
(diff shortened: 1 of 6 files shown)


scripts/build_docs.py
```diff
@@ -3,21 +3,25 @@
...
 
 from importlib.machinery import SourceFileLoader
 import os
 from subprocess import Popen
 
-docs_dir = os.path.dirname(os.path.abspath(__file__))
-version_file = os.path.join(docs_dir, os.pardir, "agentos", "version.py")
+scripts_dir = os.path.dirname(os.path.abspath(__file__))
+version_file = os.path.join(scripts_dir, os.pardir, "agentos", "version.py")
 loaded = SourceFileLoader("agentos.version", version_file).load_module()
```
andyk, line 15: Maybe it would be cleaner if we used `agentos.version` here?

scripts/build_docs.py
```diff
@@ -3,21 +3,25 @@
...
+  $ python scripts/build_docs.py
 """
 
 from importlib.machinery import SourceFileLoader
 import os
 from subprocess import Popen
 
-docs_dir = os.path.dirname(os.path.abspath(__file__))
-version_file = os.path.join(docs_dir, os.pardir, "agentos", "version.py")
+scripts_dir = os.path.dirname(os.path.abspath(__file__))
```
andyk, line 13: Do you think it's worth moving the creation of these common path variables to `shared.py`?

andyk (approved):
LGTM as is, though I left comments with a couple of ideas that you could implement if you agree or like  them, but not necessary at all.

scripts/build_docs.py
nickjalbert: sounds good!

scripts/build_docs.py
nickjalbert: I'll play around with that.   Should I just be able to `import agentos`  directly given the state of the requirements?

nickjalbert: ah looks like `agentos.__version__` works.  Is that what you were thinking?

andyk: Ha, yes, that's what I was thinking.

nickjalbert:
comments addressed and tests passing.  merging!

PR #54, opened 2021-01-23
Add github action to ensure README is updated

nickjalbert:
Adds a new build task that will yell at us if we forget to regenerate the README after changing the docs.


andyk (approved):
Brilliant!

PR #65, opened 2021-01-29
Fix flaky chatbot test

nickjalbert:
Because of the synchronization issues with the ChatBot and the threaded env, this changes the test to directly instantiate a ChatBot rather than relying on `run_agent`.

Test:

```
pytest test_all.py -rP -v -k 'test_chatbot'
```


andyk (approved):
LGTM!

Issue #67, opened 2021-01-30
Broken link in readme


andyk:
There is currently a broken link in the [install and explore](https://github.com/agentos-project/agentos#install-and-explore) section of the readme.

This is happening because `scripts/build_docs.py` doesn't parse Sphinx-specific ReStructuredText directives such as `` :doc:`doc_name` ``.

One possible fix is to see if sphinx provides an API with necessary tools to parse the file correctly (but it would have to generate a full URL including the agentos.org part)

Another is to write some parsing logic that looks for an replaces these manually.

PR #109, opened 2021-02-26
update docs/readme based on lessons learned cutting release 0.0.7

andyk:
This partially addresses #67 by just removing the offending link (since it was causing PyPI upload to fail while trying to push release 0.0.7 to PyPI). It does not guarantee that we won't accidentally use Sphinx-specific references that don't make sense in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,8 +47,8 @@ Then, create and run your first Agent::
   agentos init
   agentos run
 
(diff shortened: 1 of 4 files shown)


nickjalbert:
LGTM!

So the problem was `:doc:` is a Sphinx only directive AND it was getting pushed into the package description AND PyPI wasn't having any of it?

nickjalbert (approved):

andyk:
Yep :-)

PR #134, opened 2021-09-02
Port Acme R2D2 agent & update ACS + Registry as necessary

nickjalbert:
Here are the updates to AgentOS that have fallen out of the Acme R2D2 agent port.

## Major Changes

* `Dataset`, `Trainer`, and `EnvironmentSpec` are now core abstractions
* `cli.py` is now a very thin wrapper around the the major AOS functionality which has been moved into `runtime.py`
* Added `ready_to_initialize()` method to all core classes<sup>[1](#readytoinit)</sup>
* The runtime creates the `shared_data` dictionary and passes it to all components to enable data sharing and communication
* There is now a `agentos.parameters` object that contains the hyperparameters contributed by components.<sup>[2](#settings)</sup>
* Updated registry to contain Acme R2D2 components:
    * [acme_r2d2_policy](https://github.com/nickjalbert/r2d2)
    * [acme_r2d2_trainer](https://github.com/nickjalbert/r2d2_trainer)
    * [acme_r2d2_dataset](https://github.com/nickjalbert/reverb_dataset)
    * [cartpole](https://github.com/nickjalbert/cartpole)
* Added a test that runs through the current Acme demo: `pytest tests/test_acme_demo.py`<sup>[3](#demoscript)</sup>

## Minor Changes

* Renamed `--iterations` CLI flag to `--episodes`
* Initialization templates moved out of `cli.py` and into `agentos/templates/`
* `_run` from the private branch is now `run_agent`. This breaks the example agents.
* Renamed `.acr/` directory to `.aos/` (kept it a dot directory, though)
* Renamed `agent.ini` to `agentos.ini`
* Moved and broke up `test_all.py` to `tests/test_core.py` and `tests/test_acme_demo.py`.  We only run `tests/test_core.py` on push to Github because we don't yet have a way to automatically install requirements for the Acme demo test.

## Future Work

* Holistic documentation update
* Port example agents
* Convert `agentos.ini` to `agentos.yaml`
* Implement `Dataset.next()` in the Acme R2D2 port
* (Maybe) move `step()` and `rollout()` from `Agent` base class to either the runtime or to the template Agent class

## Footnotes

<a name="readytoinit">**1**</a>: The problem this is trying to solve: sometimes components have dependencies on each other that require one to be initialized before the other.  For example, [`R2D2Policy` needs the `environment_spec` to be in `shared_data`](https://github.com/nickjalbert/r2d2/blob/main/policy.py#L38) but that is only added when the [`CartPole` environment gets initialized](https://github.com/nickjalbert/cartpole/blob/main/cartpole.py#L17).

My current solution is that components can optionally override the `ready_to_initialize()` method and check the state of `shared_data` (or any other dependency) to determine if all pre-reqs have been met before the AOS runtime attempts to initialize them.  I'm not sold on this approach and would be happy to discuss alternatives.

<a name="settings">**2**</a>: A quick rundown on how parameters work: each component can optionally contain a `parameters.yaml` file that specifies the parameters for the component. Here's [an example from R2D2Policy](https://github.com/nickjalbert/r2d2/blob/main/parameters.yaml).  The runtime makes the parameters of each component globally available through the `agentos.parameters` object (e.g. `agentos.parameters.store_lstm_state`).  If two components attempt to define the same parameter and assign it a different value, then the runtime will throw an error.

<a name="demoscript">**3**</a>: This demo script being:

```
mkdir dev_agent
cd dev_agent
agentos init .
agentos install acme_r2d2_policy
agentos install acme_r2d2_dataset
agentos install acme_r2d2_trainer
agentos install cartpole
agentos run
agentos learn -e 100
agentos run
```


nickjalbert:
Working on getting build green.

nickjalbert:
Build is green; ready for review!

agentos/runtime.py
```diff
@@ -0,0 +1,434 @@
...
+import importlib.util
+from pathlib import Path
+import configparser
+import importlib
+import statistics
+
+
+# TODO - reimplement hz, max_iters, and threading
+def run_agent(
+    episodes,
```
andyk, line 21: `num_episodes`?

agentos/cli.py
```diff
@@ -162,114 +19,27 @@ def agentos_cmd():
     "-l",
     metavar="PACKAGE_LOCATION",
     type=click.Path(),
-    default="./.acr",
+    default="./.aos",
     help="Path to AgentOS Component Registry installation directory",
 )
 @click.option(
     "--agent-file",
```
andyk, line 26: Have you tested the `install` CLI command from outside of a directory using the `--agent-file` command? I vaguely remember trying that at some point and it failing.

agentos/templates/MLProject
```diff
@@ -0,0 +1,10 @@
+{file_header}
```
andyk, line 1: Is MLflow still a dependency given the current state of the CLI? Or did we get rid of the use of MLflow for running agents under certain conditions (which is how it used to be used)? ...in which case we can remove MLflow as a dependency and remove this template file too?

andyk (changes requested):

andyk:
Thanks for the PR. I'm excited to see this updated prototype of the registry and component system that can support a ported Acme agent! Left some questions and requests changes (mostly small I think).

andyk:
One other observation: the

agentos/__init__.py
```diff
@@ -5,16 +5,38 @@
...
     "Agent",
     "Policy",
     "Environment",
+    "EnvironmentSpec",
+    "Dataset",
+    "Trainer",
     "run_agent",
-    "rollout",
-    "rollouts",
+    "install_package",
```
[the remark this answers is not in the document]
nickjalbert: Good thought, I don't have a strong preference here. Do you?  

For now, I'll remove the functions that are currently only used in `cli.py` and `runtime.py` from the `agentos` namespace (specifically  `install_component()`, `initialize_agent_directories()`, `load_agent_from_path()`).

Note, I also renamed `install_package()` to `install_component()` to start working on regularizing the terminology used in the code base.

agentos/core.py
```diff
@@ -45,14 +52,58 @@ class Agent(MemberInitializer):
...
+            self.curr_obs, self.environment.valid_actions
+        )
+        prev_obs = self.curr_obs
+        self.curr_obs, reward, done, info = self.environment.step(action)
+        self.dataset.add(prev_obs, action, self.curr_obs, reward, done, info)
+        if done:
+            self._should_reset = True
+        return prev_obs, action, self.curr_obs, reward, done, info
+
+    def rollout(self, should_learn):
```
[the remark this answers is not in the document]
nickjalbert: Good point.  Acme actually handles this by calling their `train()` function (called [`Agent.update()`](https://github.com/deepmind/acme/blob/master/acme/environment_loop.py#L100) in their case) after *every* step in the environment and then Trainer is responsible for [deciding if/how much it will actually train during that particular call](https://github.com/deepmind/acme/blob/master/acme/agents/agent.py#L80).

I think I like the Acme behavior and will update this function to mimic it.  This would make it a convention that the Trainer gets a chance to train after every step in the environment (and it can decide to take that chance or not).  

Let me know how you feel about this design choice; happy to consider alternatives as well.

agentos/core.py
```diff
@@ -45,14 +52,58 @@ class Agent(MemberInitializer):
...
+            step_count += 1
+        if should_learn:
+            self.trainer.improve(self.dataset, self.policy)
+            prev_step_count = self.get_step_count()
+            prev_episode_count = self.get_episode_count()
+            self.save_step_count(prev_step_count + step_count)
+            self.save_episode_count(prev_episode_count + 1)
+        return step_count
 
     def advance(self):
```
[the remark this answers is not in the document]
nickjalbert: Yeah, I guess there's no standard name for the method that progresses the agent forward.  I'll rename `Agent.step()` to `Agent.advance()` to harmonize better with original AOS design and avoid aliasing with `Environment.step()`.

agentos/core.py
```diff
@@ -45,14 +52,58 @@ class Agent(MemberInitializer):
...
 
-    def learn(self):
-        """Does one iteration of training"""
-        pass
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.curr_obs = None
+        self._should_reset = True
+
+    def step(self):
```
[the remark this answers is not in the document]
nickjalbert: > Can you say more about why you pushed the stepping/rollout logic from the runtime into the agent?

My thinking was that rolling out an episode is a pretty standard thing to do with an Agent and that running a rollout requires access to all the components within the Agent (i.e. Environment, Policy, Trainer, Dataset), thus it made sense to put it as a method on Agent.

A couple of thoughts while I'm digging around here:
* It seems odd for Agent to contain (or, at least, have a pointer to) the Environment given the standard textbook depiction of the Agent/Environment loop.  This might be confusing to people coming to AOS for the first time. Is this what motivated you to try renaming Agent to Composer?  No solution from me at the moment but it did strike me as I was looking at the code again.
* Looking at some of your example agents, I can see how a separate `rollout()` can also be useful (e.g. for an evolutionary agent evaluating multiple policies).  Maybe when we port forward the example agents, we reconstitute a top-level rollout and then `Agent.rollout` becomes a wrapper around it?

> Also why did you chose the name step() for this?

Renaming to `advance()` as per my comment [here](https://github.com/agentos-project/agentos/pull/134#discussion_r703046444).

agentos/templates/MLProject
nickjalbert: Yeah, I think this and conda are both currently unused.  Removing for now although I suspect MLFlow or both will be back.

agentos/cli.py
nickjalbert: The following works for me:

```
mkdir test_agent
cd test_agent/
agentos init .
cd ..
agentos install cartpole -d test_agent/.aos/ -f test_agent/agentos.ini
cd test_agent/
agentos run
```

I'm guessing the problem is you weren't specifying both the AOS directory with `-d` **and** the `agentos.ini` with `-f`.   For now, I'll just throw an error if you specify `-d` without `-f` or vice versa.  

We could just have the `agentos.ini` specify the AOS directory path but I was thinking that you would maybe want to install an agent multiple time in different directories using the same `agentos.ini` file.  Maybe that's too much future-proofing though.

nickjalbert:
Alright @andyk, I think I've addressed all your comments.  I left a few unresolved so you can see the response and potentially give me more feedback.

Regarding your `ready_to_initialize()` comment: as discussed on the phone, I think you're right.  Not going to do anything to it right now, but we may be able to drop that when we move to hierarchical components.

agentos/cli.py
andyk: Ok, cool. I just tried your script in WSL on my Windows machine, and it works.

andyk (approved):
Thanks for your updates and responses. This LGTM. Merging now.

agentos/core.py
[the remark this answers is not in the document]
nickjalbert: :+1: giving #136 a look now!

PR #137, opened 2021-09-10
`agentos reset` and other minor functionality updates

nickjalbert:
A few small features that helped with debugging as I was finishing the Acme R2D2 port:

* `agentos reset` removes any data backing the agent (e.g. NNs backing a policy, training statistics) so you can re-train from scratch.  `agentos reset -b [backup UUID]` restores the agent backing data specified by the UUID.  I imagine the mechanisms for these might change as you get deeper into your work on hierarchical components and MLFlow.
* Added back in the `--max-iters` flag (renamed to `--max-transitions`) for `agentos run` and `agentos learn`.
* Updated the default environment (the corridor environment you get after `agentos init`) so that it will work with the Acme R2D2 agent.  For example:
```
mkdir corridor_agent
cd corridor_agent
agentos install acme_r2d2_policy
agentos install acme_r2d2_trainer
agentos install acme_r2d2_dataset
agentos run -m 200
```
* The agent prints all its hyperparameters to the console when it runs.

### DIFF
--- a/agentos/cli.py
+++ b/agentos/cli.py
@@ -85,6 +85,13 @@ _option_test_num_episodes = click.option(
     help="Number of episodes to run for performance eval.",
 )
 
+_option_max_transitions = click.option(
+    "--max-transitions",
+    "-m",
+    type=click.INT,
+    default=None,
+    help="The maximium number of transitions before truncating an episode",
+)
 _option_verbose = click.option(
     "--verbose",
     "-v",
@@ -138,6 +145,7 @@ def init(dir_names, agent_name, agentos_dir):
 @_option_test_num_episodes
 @_option_agent_file
 @_option_agentos_dir
+@_option_max_transitions
 @_option_verbose
 def learn(
     num_episodes,
@@ -145,6 +153,7 @@ def learn(
     test_num_episodes,
     agent_file,
     agentos_dir,
+    max_transitions,
     verbose,
 ):
     _check_path_exists(agentos_dir)
@@ -154,6 +163,7 @@ def learn(
         test_num_episodes=test_num_episodes,
         agent_file=agent_file,
         agentos_dir=agentos_dir,
+        max_transitions=max_transitions,
         verbose=verbose,
     )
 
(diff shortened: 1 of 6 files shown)


agentos/cli.py
```diff
@@ -131,56 +144,63 @@ def init(dir_names, agent_name, agentos_dir):
...
-# TODO - reimplement hz and max_iters
+# TODO - reimplement hz
 @agentos_cmd.command()
 @_option_num_episodes
 @_option_agent_file
 @_option_agentos_dir
+@_option_max_transitions
 @_option_verbose
-def run(num_episodes, agent_file, agentos_dir, verbose):
+def run(num_episodes, agent_file, agentos_dir, max_transitions, verbose):
```
andyk, line 183: +1 to supporting a `max_transitions` argument. What do you think of making it optional (i.e. have a default value of `None`)? If it is required, then one can't let an agent just run till its done regardless of how long it takes to achieve that, which seems like an option we want the developer to have.

agentos/core.py
```diff
@@ -66,30 +66,46 @@ def advance(self):
             self._should_reset = True
         return prev_obs, action, self.curr_obs, reward, done, info
 
-    def rollout(self, should_learn):
+    def rollout(self, should_learn, max_transitions):
```
andyk, line 69: My comment above bout making `max_transitions` optional would apply here too.

agentos/runtime.py
```diff
@@ -105,19 +101,22 @@ def learn(
     test_num_episodes,
     agent_file,
     agentos_dir,
+    max_transitions,
```
andyk, line 104: Same comment from above about making this optional.

andyk (changes requested):
Looks good. There is point I'd like to discuss, and possibly change, before we merge (see my comments in the review).

nickjalbert:
Sounds good, made max_transitions optional everywhere!

andyk (approved):
LGTM!

PR #141, opened 2021-09-16
fixes #140 - Drop .improve() and observe() from template policy

### LINKED ISSUE #140: Drop `.improve()` and `observe()` from template policy
Since we provide/support a `Trainer` class which itself has an `improve()` function, it seems cleaner to leave improving a policy up to the Trainer. So I think we should delete this stub function:

https://github.com/agentos-project/agentos/blob/0d71c3e3af81e848e2450e60cfe16a03f9b18a17/agentos/templates/policy.py#L11-L12

And since not all agents will requires an `policy.observe()` function, perhaps we should delete that from the template as well?

https://github.com/agentos-project/agentos/blob/0d71c3e3af81e848e2450e60cfe16a03f9b18a17/agentos/templates/policy.py#L14-L15

andyk:


nickjalbert:
LGTM!

PR #169, opened 2021-10-12
Fixes #161. Refactors runtime.

### LINKED ISSUE #161: AgentOS should use __init__() instead of init() when building the component graph
Currently, the initialization of each component in the component graph happens in three stages:
1. an instance is created from the component_class specified by the component spec. This of course happens using standard Python class initialization, so the component class's __init__() is called. This happens as part of`_load_component()`, at line 109 of runtime.py:
https://github.com/agentos-project/agentos/blob/0909de7a57876c10208bb6f34ccac160160adde5/agentos/runtime.py#L109
2. Then all of the the component's dependencies are recursively initialized and an attribute is created on this component to each of its dependencies. This is also part of the recursive function `_load_component()` at line 120 of runtime.py:
 https://github.com/agentos-project/agentos/blob/0909de7a57876c10208bb6f34ccac160160adde5/agentos/runtime.py#L123
3. Finally, the init() function is called on the component instance. This happens after the entire DAG of component instances is created (i.e., after the recursive `_load_component()` returns) in `_load_component_from_file()` at line 207:
https://github.com/agentos-project/agentos/blob/0909de7a57876c10208bb6f34ccac160160adde5/agentos/runtime.py#L207
and this happens in depth first order (by popping off of a stack that was created during the recursive call to `_load_component()`) so that an agent developer can use the `self.some_dependency_component` members in the `init()` function.

It is ugly that AgentOS uses both __init__() and (if it exists) `init()`, and it would be much cleaner if we could use only `__init__()`. We didn't program it this way so far because we want to allow circular dependencies and to support that requires being able to instantiate all of the instances in the DAG before calling `__init__()` on any of the instances. It is probably possible to do this by using some more advanced Python functionality (metaclasses or staticmethods maybe?).

andyk:
Fixes #161 and refactors the runtime a little bit, hopefully simplifying it and making it a bit more readable.

Adds a test case to tests/test_core that runs example_agents/sb3_agent's `evaluate` and `learn` entry points.

These changes simplify the process of loading the graph quite a bit. This also disallows cyclical references in the component DAG.

Also includes some work in progress to port some of the old example agents over to the current version of AgentOS. 

Fixes a semantic bug in `agentos.Agent` where we were still defining an `init()` function.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -114,7 +114,7 @@ def run_component_in_dir(
         print(f"Installing {req_file} with cwd {dir_name}")
         virtualenv.run(
             ["pip", "install", "-r", req_file],
-            cwd=Path(dir_name),
+            cd=Path(dir_name),
             capture=True,
         )
     for entry_point in entry_points:
@@ -123,6 +123,7 @@ def run_component_in_dir(
         virtualenv.run(args, cwd=Path(dir_name), capture=True)
 
 
+@pytest.mark.skip(reason="TODO: fix installing sb3 requirements.")
 def test_sb3_agent(virtualenv):
     agent_dir = Path(__file__).parent.parent / "example_agents" / "sb3_agent"
     run_component_in_dir(


andyk:
This has a critical bug in the new component DAG loading logic that tests/test_core.py found. I'm still fixing this bug.

PR #178, opened 2021-10-17
Add tracker component to Acme DQN agent

nickjalbert:
**Update**

Sorry, kind of dropping a big diff on you here.  One possible way to approach this is:

* Take a look at the changes in [`agentos/runtime.py`](https://github.com/agentos-project/agentos/pull/178/files#diff-e199550eafda7fb5c2728989fe07e679cd1aebc91ecd56b6a8ddfe22fc65fe8fR54) to see core updates needed to support this tracking component scheme.
* Check out any of the example agents (`example_agents/acme_r2d2/agent.py`, `example_agents/acme_dqn/agent.py`, and/or `example_agents/sb3_agent/agent.py`) to get a feel for how this particular tracker implementation affects the agent entry points.
* Take a look at the framework specific trackers (SB3: `example_agents/sb3_agent/tracker.py`, Acme: `example_agents/acme_dqn/tracker.py`) and the base tracker (`agentos/tracker.py`) to see how the integration with MLflow works.
* Run some of the example agents.

Most of the other changes are clean up and/or bookkeeping.  The high-level impact of this PR (assuming we choose to go this route) is that it gets us back to a point where we locally manage and store data to upload to the leaderboard (although we still have to decide on exactly what gets uploaded on the component implementation side).

**Original**

This prototype is a start toward some of the ideas we discussed over chat and standup the past few days; what I've done:
* Added a new `AcmeTracker` component to the Acme DQN agent port.  This tracker uses MLflow under the hood to manage an agent's state, params, and metrics.
* Updated the runtime to attach an extra `__agentos__` attribute to all components.  This attribute is a dictionary that contains the params, entry point, class name etc.  This is so the `AcmeTracker` (or any other tracking tool) doesn't need privileged access to the AOS runtime to make a run reproducible.

You can demo the DQN agent as follows (I've confirmed it learns to play Cartpole):

```bash
# Setup
cd example_agents/acme_dqn
virtualenv -p /usr/bin/python3.9 env
source env/bin/activate
pip install -r requirements.txt

# Train the agent
agentos run agent --entry-point learn --param-file parameters.yaml -Pnum_episodes=300

# Evaluate the agent
agentos run agent --entry-point evaluate --param-file parameters.yaml

# Reset the agent: bye bye backing state
agentos run agent --entry-point reset --param-file parameters.yaml
```

Note that the `AcmeTracker` tracks interesting metrics across multiple different runs (i.e. number of episodes and transitions the agent has been trained on over its whole life) similar the previous "core data" tooling.  For example, look at this output from an evaluation run:

```bash
Benchmark results after 20 episodes:
        Benchmarked agent was trained on 114537 transitions over 820 episodes
        Max steps over 20 trials: 500
        Mean steps over 20 trials: 331.9
        Median steps over 20 trials: 266.5
        Min steps over 20 trials: 208
```

You can also browse the backing run data using the standard `mlflow ui`.

**Higher level**

With this approach, experiment management is just another component.  I imagine that AgentOS could provide trackers for many of the major frameworks (e.g. `AcmeTracker`, `SB3Tracker`, etc) to make it easy to port agents into AgentOS and then integrate them with the leaderboard and registry.

Further, I imagine we could then have an open spec describing how to upload a run to the leaderboard.  Then the built-in AgentOS trackers could provide automated tooling to handle it for you (e.g. `agentos run tracker --entry-point upload`), but a developer would be free to create their own toolchain as well.

**Potential next steps**

* Any feedback on this particular approach to tracking and reproducibility?
* We should discuss further your idea of capturing each MLflow run as its own component.  I think there are trade-offs here (i.e. it makes the component model more complex) but could be powerful.  I imagine this could work together with something like `AcmeTracker` or as an alternative to it.
*  *see comment below* ~~I want to see how easy it is to incorporate `AcmeTracker` into the R2D2 agent~~
* *see comment below* ~~I want to see how much tracking functionality can be shared between `AcmeTracker` and `SB3Tracker`~~

### DIFF
--- a/agentos/tracker.py
+++ b/agentos/tracker.py
@@ -1,32 +1,11 @@
 import mlflow
+from mlflow.entities import Run
 import yaml
 import tempfile
 import shutil
 import statistics
+from typing import List
 from pathlib import Path
-
-
-class EvaluateRunManager:
-    def __init__(self, tracker):
-        self.tracker = tracker
-
-    def __enter__(self):
-        self.tracker.start_evaluate_run()
-
-    def __exit__(self, type, value, traceback):
-        self.tracker.print_results()
-
-
-class LearnRunManager:
-    def __init__(self, tracker):
-        self.tracker = tracker
-
-    def __enter__(self):
-        self.tracker.start_learn_run()
-
-    def __exit__(self, type, value, traceback):
-        self.tracker.log_learn_run_metrics()
-        self.tracker.print_results()
 
 
 class AgentTracker:
@@ -67,7 +46,7 @@ class AgentTracker:
             "parameter_file.yaml", self.__agentos__["fully_qualified_params"]
         )
 
-    def log_data_as_artifact(self, name, data):
+    def log_data_as_artifact(self, name: str, data: dict):
         dir_path = Path(tempfile.mkdtemp())
         artifact_path = dir_path / name
         with open(artifact_path, "w") as file_out:
@@ -82,13 +61,13 @@ class AgentTracker:
         self.log_episode_count(data["episodes"])
         self.log_step_count(data["steps"])
 
-    def log_episode_count(self, count):
+    def log_episode_count(self, count: int):
         mlflow.log_metric(self.EPISODE_KEY, count)
 
-    def log_step_count(self, count):
+    def log_step_count(self, count: int):
         mlflow.log_metric(self.STEP_KEY, count)
 
-    def get_training_info(self):
+    def get_training_info(self) -> (int, int):
         runs = self._get_all_runs()
         total_episodes = 0
         total_steps = 0
@@ -100,7 +79,7 @@ class AgentTracker:
                 total_steps += int(run.data.metrics.get(self.STEP_KEY, 0))
         return total_episodes, total_steps
 
-    def _get_all_runs(self, respect_reset=True):
+    def _get_all_runs(self, respect_reset: bool = True) -> List[Run]:
         assert mlflow.active_run() is not None
         run_infos = mlflow.list_run_infos(
             experiment_id=self.MLFLOW_EXPERIMENT_ID,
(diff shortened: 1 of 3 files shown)


nickjalbert:
I converted `acme_r2d2` agent to use `AcmeTracker` (in e357b97).  It was pretty straightforward and simplified the R2D2 agent a bit.  A few notes:

* The main meat of the change happened in [`R2D2Agent`](https://github.com/agentos-project/agentos/pull/178/files#diff-53b88749d4f6981ba6140f57b566e50f25c71192e6aa4c96c1c3e67af5c5c59bR4).  The agent class no longer inherits from `agentos.Agent` and looks very similar to [`AcmeDQNAgent`](https://github.com/agentos-project/agentos/pull/178/files#diff-04e090704b1a5031afbdcd667edc3cf159cd9408c62bd6f004a8969d48da8028R6).  I added a few methods to `R2D2Agent` so that I could pass it directly into Acme's [EnvironmentLoop](https://github.com/deepmind/acme/blob/93a6a15b8c7b1d6a9c8099b9f2e9c693ab11a41e/acme/environment_loop.py#L32).
* I directly reused the `AcmeTracker` and eliminated the duplicated `CartPole` implementation by pointing the agent spec file in `acme_r2d2` to `acme_dqn`'s tracker and environment.
* I converted the R2D2 agent spec to use [yaml anchors and tags](https://github.com/agentos-project/agentos/pull/178/files#diff-76fb3e2061135b03112f821e569bad47fbd0061fff858a72c4fc885a9406017fR1) instead of `__global__`

andyk:
> Made a `BaseTracker` class in `agentos/tracker.py` (which can probably replace `agentos/tracking.py`) for sharing functionality.

I support removing `tracking.py` as part of this PR then!

PR #186, opened 2021-10-27
Component test

nickjalbert:
This is not for merge, just a demo of using the component system to instantiate two components that use different versions of numpy.  

Here `ComponentTestAgent` has two dependencies: `NumpyTestA` and `NumpyTestB`.  The agent calls the `check_numpy_version()` method on each dependency that checks if:
*  `NumpyTestA` is running against numpy version 1.21.3 
*  `NumpyTestB` is running against numpy version 1.20.0

We also check that a method that was added between these two versions of numpy exists in `NumpyTestA`'s version but not in `NumpyTestB`'s version.

The tooling in `runtime.py` is pretty straightforward, it installs the declared requirements to a component-specific cache directory and imports it within that environment.

You can demo by running the following in `component_test/`:
```
agentos run agent
```

### DIFF
--- a/agentos/runtime.py
+++ b/agentos/runtime.py
@@ -321,7 +321,7 @@ def _handle_env_manipulation(section):
             subprocess.run(cmd)
         sys.path.insert(0, str(env_path))
     try:
-        yield module_file
+        yield module_file, bool(req_list)
     finally:
         sys.path.pop()
         if req_list:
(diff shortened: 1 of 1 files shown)


nickjalbert:
I'm going to close this for now and create a new PR when it's more fully baked.

PR #187, opened 2021-11-02
Refactor local Component API (Components v3)

nickjalbert:
Here are the changes that I demo'd yesterday (see presentation [WIP: Components V3](https://docs.google.com/presentation/d/1NHeUml3hPQ82v3a4T3Id4wpSPMqaZeaxIvSSahbrylk/edit#slide=id.gfcf66cf896_0_168)).  

At a high-level, this PR splits out `Component` into its own file and refactors its API.  One change from the above presentation is that I renamed `Registry` to `ComponentNamespace`.  Also note that this does **not** include the component dependency isolation work found in #186.

All example agents are ported; they should run as usual. To run the "REPL-ready demo":

```
python agentos/component.py
```

**Question**: what do you think about folding the parameter file into the `agentos.yaml`?  For example, the resulting file for Acme DQN could look something like:

```
repos:
    local_dir:
        type: local
        path: .


components:
    agent:
        repo: local_dir
        file_path: ./agent.py
        class_name: AcmeDQNAgent
        dependencies:
            network: network
            environment: environment
            tracker: tracker
        parameters:
            __init__:
                discount: .99
                batch_size: 10
                samples_per_insert: 2
                min_replay_size: 10
            evaluate:
                num_episodes: 100
            learn:
                num_episodes: 200

    network:
        repo: local_dir
        file_path: ./network.py
        class_name: AcmeDQNNetwork
        dependencies:
            environment: environment
            tracker: tracker

    environment:
        repo: local_dir
        file_path: ./environment.py
        class_name: CartPole
        parameters:
            __init__:
                discount: .99

    tracker:
        repo: local_dir
        file_path: ./tracker.py
        class_name: AcmeTracker
```


agentos/component.py
```diff
@@ -0,0 +1,252 @@
...
+
+    # Instantiate a SimpleAgent and run reset_env() method
+    agent = agent_component.get_instance()
+    agent.reset_env()
+
+    # Instantiate a ComponentNamespace
+    ns = ComponentNamespace()
+
+    # Add Components to the ComponentNamespace
+    ns.add_component("agent", agent_component)
```
andyk, line 246: As we discussed today, can we make the Namespace API internal and not user facing for introductory level usage?

Basically, can we introduce the Component abstraction (and specifically the class and CLI tooling for working with Components) as the main fundamental abstraction and have as much functionality live in the Component abstraction as possible?

For now, perhaps it is ok to have a `Repo` abstraction, and that is certainly how we have been thinking about it in all of our discussions so far. But **maaaaybe** we an get away with not having a separate Repo abstraction afterall? Or at least having it be a lower level abstraction (like namespaces) that beginners wouldn't have to learn to get going with Component Programming.

agentos/component.py
```diff
@@ -0,0 +1,252 @@
...
+
+        def reset_env(self):
+            self.env.reset()
+
+    class SimpleEnvironment:
+        def reset(self):
+            print("SimpleEnvironment.reset() called")
+
+    # Generate Components from Classes
+    agent_component = Component.get_from_class(SimpleAgent)
```
andyk, line 232: As we discussed today, IMO, the easiest to learn interface would be my_component = Component(), and then we could essentially have `__init__()` be an alias for `get_from_class()`

andyk (changes requested):

andyk:
I have a few thoughts about why it might make sense to keep the params file separate from the component spec that I didn't write up yet too. Can discuss live or else I'll try to write up my thoughts in next day or two.

agentos/component.py
```diff
@@ -0,0 +1,252 @@
...
+import sys
+import yaml
+import typing
+
+# https://stackoverflow.com/a/44331733
+T = typing.TypeVar("T")
+
+
+class Repo:
+    LOCAL_TYPE = "local"
```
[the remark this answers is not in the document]
nickjalbert: Yeah, I'm using it as an Enum there, and it probably makes sense to create a real Enum.  

In terms of style,  I think you could read the [pep-8 constants entry](https://www.python.org/dev/peps/pep-0008/#constants) as suggesting all caps for constants anywhere (and "usually" they're defined at the module-level). 

Seems to appear in the wild in some places, e.g. the examples in the [Enum docs](https://docs.python.org/3/library/enum.html):

enum

Or in the [Django docs](https://docs.djangoproject.com/en/3.2/topics/db/models/):

django
 
Not committed to this style choice if you have a different preference!

agentos/component.py
nickjalbert: Agreed on your first point about making the Namespace API internal.

And, yeah, right now `Repo` doesn't feel super useful, maybe I'll rip it out and see if that makes things ugly.

nickjalbert:
Alright, nits addressed!  Left `Repo` in for now but refactored the `Component` as we discussed.  I also broke out the demo that was at the bottom of `component.py` into its own test.

PR #190, opened 2021-11-10
migrates aos_web from nickjalbert's github to main repo

andyk:
Copies the contents of https://github.com/nickjalbert/aos_web into a directory called `web`


andyk:
Would it make more sense to have the web dir live under the agentos dir?

PR #199, opened 2021-11-15
Ability to pin a version of a spec file

nickjalbert:
This PR adds functionality to take a Component DAG and pin the version of each Component to a publicly available version on GitHub.  The resulting spec can then be run on any machine where AgentOS is installed.

## Demo

```bash
cd example_agents/sb3_agent/

# Run the local version of the SB3 PPO example agent
agentos run agent

# Pin the version of SB3 PPO example agent
agentos pin agent

# Copy the updated agentos.yaml into another directory
mkdir ~/demo
cp agentos.yaml ~/demo/

# Run the agent from the updated spec that has AgentOS pull the Component tree from GitHub
cd ~/demo
agentos run agent
```

## Caveats

Before pinning, we require that each Component in the dependency tree:
* Be in a repo (not necessarily the same repo) whose `origin` is a GitHub URL
* That the local repo branch has a corresponding branch on `origin` and both branches are at the same commit
* That there are no uncommitted changes in the repo

### DIFF
--- a/agentos/cli.py
+++ b/agentos/cli.py
@@ -154,12 +154,12 @@ def run(
 @agentos_cmd.command()
 @_arg_component_name
 @_option_component_spec_file
-def pin(component_name, component_spec_file):
+def freeze(component_name, component_spec_file):
     """
     Creates a version of ``component_spec_file`` for Component
-    ``component_name`` where all Components in the dependency tree are pinned
-    to a specific git commit.  The resulting ``component_spec_file`` can be run
-    on any machine with AgentOS installed.
+    ``component_name`` where all Components in the dependency tree are
+    associated with a specific git commit.  The resulting
+    ``component_spec_file`` can be run on any machine with AgentOS installed.
 
     The requirements for pinning a Component spec are as follows:
         * All Components in the dependency tree must be in git repos
@@ -169,10 +169,8 @@ def pin(component_name, component_spec_file):
         * There are no uncommitted changes in the local repo
     """
     component = Component.get_from_yaml(component_name, component_spec_file)
-    pinned_spec = component.get_pinned_component_spec()
-    with open(component_spec_file, "w") as file_out:
-        yaml.dump(pinned_spec, file_out)
-    print(f"Pinned spec written to {Path(component_spec_file).absolute()}")
+    frozen_spec = component.get_frozen_component_spec()
+    print(yaml.dump(frozen_spec))
 
 
 # Copied from https://github.com/mlflow/mlflow/blob/3958cdf9664ade34ebcf5960bee215c80efae992/mlflow/cli.py#L188 # noqa: E501
--- a/agentos/cli.py
+++ b/agentos/cli.py
@@ -4,6 +4,7 @@ The CLI allows creation of a simple template agent.
 """
 import os
 import sys
+import yaml
 import click
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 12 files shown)


andyk:
Let's call it "freeze" instead of "pin" so folks familiar with PIP will get it faster.

agentos/component.py
```diff
@@ -287,3 +322,7 @@ def _no_track_run(self, fn_name: str, params: ParameterSet):
             yield
         finally:
             pass
+
+
+def get_full_path(repo: Repo, identifier: ComponentIdentifier, file_path: str):
```
andyk, line 327: Could this be an `@staticmethod` on the Component class?

It feels a little funny hanging out down here at the module-level.

agentos/repo.py
```diff
@@ -62,6 +67,7 @@ class GitHubRepo(Repo):
     def __init__(self, name: str, url: str):
         self.name = name
         self.type = RepoType.GITHUB
+        url = url.replace("git@github.com:", "https://github.com/")
```
andyk, line 70: Why do we enforce this? Does Dulwich only work with HTTPS and not ssh?

andyk (approved):
LGTM.

I left some comments in the code, one in the main PR thread (rename pin to freeze) and one below.

Since `Component.get_from_class()`, `Component.get_from_repo`, `Component._parse_components()`, and `Repo.from_spec()` aren't using the `cls` method (and in fact `get_from_repo()` creates a local variable called `cls` which is slightly confusing since it is unrelated to the method's `cls` argument), can they be `@staticmethod`s instead.

And then, if you change the currently module-level `get_full_path` to be a @staticmethod as well, then you'd have one static method (`get_from_repo`) calling another static method (`get_full_path`) and also an instance method (`get_full_path`) calling the same static method (`get_full_path`).

agentos/component.py
nickjalbert: This weirdness is the result of a somewhat surprising Python behavior that I bumped into.  Consider the following class:

```Python
class A:
    @staticmethod
    def foo():
        print('static method')

    def foo(self):
        print('instance method')
```

I ran the following code:

```python
A.foo()
a = A()
a.foo()
```

and expected to see:

```bash
$ python test.py
static method
instance method
```

but what I ended up seeing was:

```bash
$ python test.py
Traceback (most recent call last):
  File "/home/nickj/agentos/web/test.py", line 11, in <module>
    A.foo()
TypeError: foo() missing 1 required positional argument: 'self'
```

It makes sense if you think about how `foo` only has one entry in the class `__dict__` so it can't be both a static and instance method but that's not how I was originally thinking about it (see also the [StackOverflow answer here](https://stackoverflow.com/a/20533442)).

There are other options (such as dynamically creating an instance method in the `__init__` method).  There's also a chance that this could be an instance method on `Repo` or `ComponentIdentifier`.  I'll see if I can find a better spot to put this function.

agentos/cli.py
```diff
@@ -146,6 +151,30 @@ def run(
     component.run(entry_point, parameters)
 
 
+@agentos_cmd.command()
+@_arg_component_name
+@_option_component_spec_file
+def pin(component_name, component_spec_file):
```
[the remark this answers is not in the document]
nickjalbert: I like mimicking the behavior of `pip freeze` here!

agentos/repo.py
nickjalbert: The difference, I believe, is the auth flow that git uses.  On my machine:
* Cloning a `git@github.com` repo makes me enter the password on the key I use to auth with github (even though I'm just trying to read and not write the repo).  The clone process just hangs until I unlock the key.
* A `https://github.com/` repo, on the other hand, can clone a repo without any user input.

I'm not sure if this is universally true or just particular to my setup.  For now I'll add a comment to note this and keep a mental note to make the git interaction process more robust overall.

nickjalbert:
Thanks for the review!  I've addressed the nits and added a test for the freezing functionality.  Will merge once we're green.

PR #200, opened 2021-11-17
Component, ComponentDependency, and Repo models

nickjalbert:
Here are the Component, ComponentDependency, and Repo model changes from the local branch we've both been working against.


web/registry/models.py
```diff
@@ -9,64 +9,54 @@ class Meta:
...
 
 
+class ComponentDependency(TimeStampedModel):
+    depender = models.ForeignKey(
+        "Component", on_delete=models.CASCADE, related_name="depender_set"
+    )
+    dependee = models.ForeignKey(
+        "Component", on_delete=models.CASCADE, related_name="dependee_set"
+    )
+    alias = models.TextField()
```
andyk, line 19: You don't necessarily have to change this since new users of Components probably won't be looking at this code anyway, but in general I feel like calling this "attribute_name" instead of "alias" would introduce one fewer new concepts to developer learning component programming for the first time, since they will already know what Python attributes are.

andyk (approved):
LGTM. One nit-level comment

nickjalbert:
Thanks again for the review @andyk!  I renamed `alias` to `attribute_name` in the code base.  

You will need to run the new migration if you're running the web server against master after this merges (`./manage.py migrate` should do the trick).

PR #208, opened 2021-12-15
Reworks registry to now handle component_specs

andyk:


andyk:
@nickjalbert I believe this addresses all of the comments you gave me and is ready for review. All tests pass except `gh_sb3_agent` which needs to have updates pushed to GitHub to pass.

Conceptually, the Registry (and its subclasses) is responsible for storing Specs (Component, Repo, Run, etc.) and loading/saving them from backing stores. Specs are Dicts that now have explicit types defined for them in `specs.py`. All specs are types which are composed out of `strs` and `dicts`.

Component is responsible for going back and forth from specs to component objects.

nickjalbert (approved):
LGTM!  A few minor comments, feel free to ignore.  

I ported the failing registry tests (they only run on linux machines because of the Reverb requirements which is why you weren't seeing them fail earlier), the `gh_sb3_agent`, and the web tests.  Hopefully build will be green!

nickjalbert:
Ran into an issue with `test_registry_from_dict()`:  it tried to run the SB3 agent, but the GitHub environment doesn't have the `stable_baselines3` package installed in it.  My fix was to add another example agent (`example_agents/random`) with no external requirements for testing things like this.

andyk:
> Ran into an issue with `test_registry_from_dict()`: it tried to run the SB3 agent, but the GitHub environment doesn't have the `stable_baselines3` package installed in it. My fix was to add another example agent (`example_agents/random`) with no external requirements for testing things like this.

Oh right. I knew that test would break things because I didn't require it to use a virtual machine that set up the dependencies. Thanks for coming up with that solution!

PR #212, opened 2022-01-03
Run fixes and tweaks

nickjalbert:
A few small fixes found during my review of https://github.com/nickjalbert/agentos/pull/8

Highlights:

* Change `Repo.to_spec()` to be a two-level dict of form `{repo_name: {...repo_info...}}` instead of one-level dict of form `{name: repo_name, ...}`.  This harmonizes it with `Component.to_spec()`.  I think this was what `Registry` was expecting because otherwise the `Registry.add_repo_spec()` implementation would probably break in a surprising way if we had multiple repos.
* Port `agentos status` command to the new registry implementation
  * TODO: add some tests for this command
* Update the Run tracking code that previously froze the component spec to now freeze the registry which includes the root component
* Handle a new way for the porcelain-git interface logic to fail
* Update the web dependency on AgentOS

### DIFF
--- a/agentos/cli.py
+++ b/agentos/cli.py
@@ -166,7 +166,7 @@ def run(
 @agentos_cmd.command()
 @_arg_optional_entity_id
 @_option_registry_file
-def status(entity_id, component_spec_file):
+def status(entity_id, registry_file):
     """
     ENTITY_ID can be a Component name or a Run ID.
     """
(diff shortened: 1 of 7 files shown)


andyk:
Thanks for fixings the Repo.to_spec issue I missed.

Regarding the nested structure of Component and (as of this PR) Repo, I'm not super happy with the way I structured them as nested dicts. I even made `Registry.get_component_spec()` return a flattened spec by default (overridable by passing in `flattened=True`). My main complaint is that it currently is cumbersome to manage these nested dicts.

I think a decent longer term solution we discussed in discord is to use NamedTuples instead, which have the nice attribute access syntax.

nickjalbert:
Yeah!  Agree the nested dicts are unwieldy; NamedTuple seems like it could be a good alternative with the drawback that then serialization requires some more work, right?

How does this PR look to you?  Ready for merge or are there some fixes you'd like?

PR #214, opened 2022-01-05
Rework the tests to avoid venv creation and installation

nickjalbert:
Let's discuss, not sure how you'll feel about this change.

In this PR I simplified the testing infrastructure so it no longer installs the requirements for the example agents in a virtualenv.  This allows us to avoid running agents via subprocesses.  

Instead, we install all the example agent requirements as a step in the github build (with the caveat that we will break this out into different build actions once we have example agents with incompatible requirements).  Three benefits of this change:

1. Simplified reasoning about the environment our tests run in.  See, for example, how we no longer need to [exec a string](https://github.com/agentos-project/agentos/compare/master...nickjalbert:nj_test_revamp?expand=1#diff-0abc741bfcebe9debcf5eb70d01e66db1ccb9036877f195680d25bf73e9e9083L70) in `test_registry.py`

2. This should speed up tests a bit on the github side because we install requirements in one go (from about ~10min in #212 to ~6min here)

3. Stack traces when running the example agents should be cleaner because we're no longer going through subprocess (although, I don't love how verbose pytest is in general).  Some tests still use subprocess and we should consider porting them to use the [click CliRunner](https://click.palletsprojects.com/en/8.0.x/testing/): 

new-error


andyk:
I like the idea a lot! If I understand this correctly, we have to now do something similar in a virtual env on our local machines if we want to run the tests locally, is that right?

If so, we should probably add a script that searches through the example_agents dirs and finds any `requirements.txt` files and merges them into a single large `example-agents-requirements.txt` and then calls pip install using that. And we probably should also update the docs (which will update the README) to reflect the new process for successfully running tests, which now includes installing all of the example agent requirements.

nickjalbert:
@andyk reworking this based on feedback now.  Got myself stuck in a bit of dependency hell, but optimistic :P

nickjalbert:
As discussed in sync, I reworked this so now dev-requirements.txt uses `-r` to install the reqs for the example_agents used in testing.  Also did a number of other cleanup tasks around the tests (updated supported Python versions to 3.7-3.9, removed all usage of subprocess in the tests, updated some packages, etc).  Build green so merging!

PR #219, opened 2022-01-12
Install CPU torch for the example agents

nickjalbert:
Minor update to RLlib agent requirements.txt to use the CPU version of torch (this satisfies the SB3 torch requirements as well so we don't get the CUDA errors).


nickjalbert:
Alright @andyk, I:

* Created a requirements installation script (`python install_requirements.py`) like you proposed
* Confirmed I can setup a new environment from scratch using the script that passes all tests on my local box 
* Confirmed all tests are passing on GH (updated the workflow to use the install script)
* Updated the README
* Updated the requirements that dependabot keeps issuing PRs for

I'm going to merge this update in.  Give it a try when you get a chance on your envs and let me know if you run into any problems!

PR #227, opened 2022-01-29
WIP (don't merge yet) RunManager -> Run, deeper Registry integration

andyk:
Creating this WIP PR as I work through the last few tests.


andyk:
Tests are now green, but before we merge, we probably want to convert this to a squashed commit and open a new PR for that since many of my intermediate commits were simply checkpoints for myself and don't have any other historical use except to muddy the git history.

andyk:
Ok @nickjalbert I've squashed all those silly "WIP" commits into one (very large) commit 2281a42 and tests are still passing so this is Ready for a review!!

I've also added a commit with the partial progress on porting the WebRegistry integration (which updates the Django webapp registry to support pushing and fetching repos and components to/from a WebRegistry).

agentos/component_run.py
```diff
@@ -0,0 +1,227 @@
...
+        super().__init__(
+            experiment_id=experiment_id, existing_run_id=existing_run_id
+        )
+        assert not (
+            run_command and existing_run_id
+        ), "`run_command` cannot be passed with `existing_run_id`."
+        self._run_command = None
+        if run_command:
+            self._run_command = run_command
+            self.log_run_command(run_command)
```
nickjalbert, line 79: It seems (maybe) a little odd that we assign `run_command` to `self._run_command` and then log it by passing in the `run_command` as an argument (instead of just using `self._run_command` in `log_run_command()` function).  

Is `log_run_command()` used anywhere else such that it needs the `run_command` argument?

agentos/component_run.py
```diff
@@ -0,0 +1,227 @@
...
+        serialization format.
+
+        :param ret_val: The Python object returned by this Run to be logged.
+        :param format: Valid values are 'pickle, 'json', or 'yaml'.
+        """
+        assert (
+            not self._return_value
+        ), "return_value has already been logged and can only be logged once."
+        self._return_value = ret_val
+        filename_base = self.identifier + "-return_value"
```
nickjalbert, line 187: It might be safer to create a file in a temp directory (as done [here](https://github.com/agentos-project/agentos/blob/7d13480e01bb1aff8a870d864ba9d1039af470fc/agentos/run.py#L270)) to avoid any surprise overwrites/removes of files that already exist in the working dir.

agentos/registry.py
```diff
@@ -226,13 +274,35 @@ def get_component_specs(
...
         return self._registry["components"]
 
-    def get_repo_spec(self, repo_id: str) -> "RepoSpec":
-        return self._registry["repos"][repo_id]
-
-    def get_run_spec(self, run_id: str) -> Dict:
-        return self._registry["runs"][run_id]
-
-    def get_registries(self) -> List[Registry]:
+    def get_repo_spec(self, repo_id: str, flatten: bool = False) -> "RepoSpec":
```
nickjalbert, line 277: it looks like the `get_*_spec()` functions are all pretty much the same.  Would it make sense to dedup the functionality?

agentos/component.py
```diff
@@ -102,36 +127,69 @@ def from_registry(
...
     ) -> "Component":
         name = name if name else managed_cls.__name__
-        return Component(
+        if (
+            managed_cls.__module__ == "__main__"
+        ):  # handle classes defined in REPL.
+            repo = LocalRepo(name)
+            src_file = repo.get_local_repo_dir() / f"{name}.py"
+            with open(src_file, "w") as f:
+                f.write(dill_getsource(managed_cls))
```
nickjalbert, line 152: Is there anyway this write unexpectedly overwrites something not specifically created/controlled by PCS?  Maybe we should check that `src_file` does not exist (if that actually is an invariant)?

agentos/component.py
```diff
@@ -167,29 +226,66 @@ def get_default_entry_point(self):
...
-            instance = (
-                instance if instance else self.get_instance(params=params)
+        entry_point: str,
+        params: Union[ParameterSet, Dict] = None,
+        publish_to: Registry = None,
+        log_return_value: bool = True,
+        return_value_log_format: str = "pickle",
+    ) -> Run:
+        """
+        Run the specified entry point a new instance of this components
```
nickjalbert, line 236: nit: "components" -> "Component's"

nickjalbert (approved):
Lots of little nits, feel free to ignore. Overall looks great to me; I'm excited to get this in! 

The biggest things (IMO) are:
* The weird MLflow error that occurs when an uncaught exception gets raised during a run
* The Acme agents producing an unexpected run structure and being unable to load in a previously-trained model
* We should be very careful when writing/removing files/dirs to not overwrite/remove things not owned by PCS.  I think it's worth programming quite defensively around this.  A few of my comments highlight spots where we should be using temp dirs or doing existence checks.

Nice work!

agentos/component.py
```diff
@@ -35,34 +39,52 @@ def __init__(
         identifier: "Component.Identifier",
         class_name: str,
         file_path: str,
+        instantiate: Optional[bool] = True,
```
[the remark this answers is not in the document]
andyk: changed to `bool`

agentos/agent_run.py
```diff
@@ -36,61 +55,51 @@ class AgentRunManager:
     AGENT_NAME_KEY = "agent_name"
     ENV_NAME_KEY = "environment_name"
 
-    def __init__(self, *args, **kwargs):
+    def __init__(
+        self,
+        run_type: str,
+        parent_run: str = None,
```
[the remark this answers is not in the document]
andyk: yep. Changed. I should probably run mypy again. That would catch all of these type errors.

example_agents/sb3_agent/components.yaml
```diff
@@ -5,20 +5,21 @@ repos:
 
 
 components:
-    agent:
+    sb3_agent:
```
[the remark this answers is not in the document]
andyk: I had changed it because when working with multiple example_agents in various tests, it was a little confusing to have all of the example agents use the same name for their components. It also was an opportunity to exercise the code path where the agent component being run has a different name than the default (and thus the user is required to specify the agent name explicitly).

I don't feel strongly about it. Happy to change it back to `agent` if you have a preference. For now I'll update the README.

andyk:
This PR leaves `.../sb3_agent> agentos publish-run` broken. I didn't catch it because it doesn't have a test case yet. I created #229 to be sure we have sufficient testing coverage.

example_agents/sb3_agent_components/agent.py
```diff
@@ -0,0 +1,43 @@
+# This file was auto-generated by `agentos init` on Aug 09, 2021 09:12:37.
+import agentos
+
+
+# A high level port of the SB3 PPO algorithm.
+class SB3PPOAgent(agentos.Agent):
```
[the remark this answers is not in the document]
andyk: No I added this agent to source control accidentally. Removing it.

agentos/component_run.py
andyk: Yeah, this is confusing. I moved the assignment into the function and the function `set_and_log_run_command`.

agentos/component_run.py
andyk: Good catch! Done.

agentos/registry.py
andyk: Done.

agentos/component.py
andyk: Good point! Updated to fail if the file exists already.

Also, note that this uses a `LocalRepo`, which takes a `local_dir` argument. The way we use that here (line 149) we end up using the default value for `local_dir` which is the following:
```python
local_dir = "./.pcs_local_repo"
```

...so hopefully using this folder provides enough of a unique namespace that we won't have a lot of failures due to encountering other files the dev has around.

agentos/component.py
andyk: Done

agentos/component.py
andyk: BTW, I'm not super happy with this...

See comment here in `repo.py`:

https://github.com/agentos-project/agentos/blob/aecb582389b83f68f394e4ac43f1631975137170/agentos/repo.py#L323-L329

tests/test_web_registry.py
```diff
@@ -0,0 +1,60 @@
...
+    assert flat_repo_spec["url"] == (
+        "https://github.com/agentos-project/agentos.git"
+    )
+
+    nested_repo_spec = web_registry.get_repo_spec("AgentOSRepo")
+    assert nested_repo_spec["AgentOSRepo"]["type"] == "github"
+    repo = Repo.from_spec(nested_repo_spec)
+    assert repo.identifier == "AgentOSRepo"
+
+    # TODO: Finish writing this test and porting WebRegistry funtionality to
```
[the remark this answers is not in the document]
andyk: Yep! It's still active WIP! ETA O(days)

andyk (commented):
Ok I think that now I have responded to all of your comments @nickjalbert and the code changes are in #234

PR #228, opened 2022-02-02
`agentos init` creates agent with NO pip dependencies on agentos library

nickjalbert:
Let's discuss; will close if we decide to go another way.  

My thinking is that we actually want two different "default" agents to be generated by `agentos init`:

* One compatible with and using *just* PCS-layer features
* One that uses PCS- and AgentOS-layer features

This PR changes the default agent created by `agentos init` to be the former.  Specifically, we don't rely on any AgentOS library functionality (i.e. we don't call `import agentos` in any of the Components).   This isolates the `agentos init` agent from changes in the currently-pretty-unstable AgentOS layer.

Once we split out PCS and AgentOS, I imagine this default agent could go with the PCS layer as a demo of a very basic componentized project (perhaps the result of `pcs init` instead of `agentos init`) and we can create a new default agent for `agentos init` that utilizes the AgentOS layer for (hopefully) significant savings in implementation plumbing.

To demo:
```bash
mkdir demo
cd demo
agentos init .
agentos run agent
agentos run agent --entry-point learn
agentos run agent --entry-point learn -P num_episodes=20
```

This also updates the agent in `example_agents/random` to exactly be this default agent.


nickjalbert:
The motivation for this was the difficulty of specifying requirements for the default agent when it depends on AgentOS libraries that aren't in an official release yet.  However, I realized that it could be useful to have the default agent less tightly-coupled to the AgentOS libraries in general as we're breaking out PCS and that this was a pretty self-contained chunk of work.

It's also nice that everything is explicitly in the `init`'d directory, rather than having to follow the inheritance chain into the AgentOS libraries to see what the default agent is doing.

andyk:
This needs to be updated to resolve the new conflicts that were introduced when I just merged #227.

nickjalbert:
@andyk fixed conflicts; ready for review!

andyk:
Excellent! I'll review it tomorrow!

andyk (approved):
LGTM. You can merge as is if you want, or I have one idea that you can optionally implement if you like it: remove the no-op trainer. While I'm not 100% convinced the default agent will be better the no-op trainer removed; it will definitely simpler, and it's a little strange to have the no-op in there.

Bigger picture, I like the idea of having a `pcs init` in addition to (or instead of) `agentos init`. Though perhaps instead of creating new stub py files `pcs init` would take a py file as input and generate a registry file for it containing a component spec (and all transitive dependency specs) for that python file.

nickjalbert:
Sounds good; I think dropping the trainer seems reasonable.  Will fix and then merge!

PR #230, opened 2022-02-03
Allow easy swapping of target branch for tests that use code from github

andyk:
Since some of our tests fetch source code from the agentos-project GitHub repo, it is helpful to be able to have that code depend on a branch that is easy to change in one place.

This will allow PR branches to easily use a `test_staging` branch and then after the PR is approved -- right before it is merged -- we can push update the `test_prod` branch to match the `test_staging` branch and change the code to use the `test_prod` branch.

I don't love the solution, but it seems much better than what we had before which is tests that depend on various static branches that we then have to hunt down an push updates to.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -155,6 +155,7 @@ repo to use when testing.
 
 If you make changes to code that is fetched from github for use by tests, then
 please follow this process for your PR:
+
 1. While doing development, change the ``TESTING_GITHUB_REPO`` and/or
    ``TESTING_BRANCH_NAME`` global variables in ``agentos/utils.py``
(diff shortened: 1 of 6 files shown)


nickjalbert (approved):
LGTM modulo whatever lint is complaining about!  Could you add a note in the README.md so we can remember how to update staging/prod next time we need to?

andyk:
Ok @nickjalbert I updated the docs/README, fixed the lint error, and also made the github repo URL a variable in addition to the branch name. If test are green I'm going to merge.

nickjalbert:
🎉

PR #232, opened 2022-02-04
Move web registry integration test to web/

nickjalbert:
This ended up being pretty straightforward.  You can now run the integration test as part of the whole web test suite:

```bash
cd web/
./manage.py test
```

or on its own:

```bash
cd web/
./manage.py test registry.tests.test_integration.WebRegistryIntegrationTestCases
```

The last few statements in the test weren't passing before I ported it, so I assumed they were works-in-progress and commented them out.

Also, I converted the raw `assert` statements to use [`self.assertEqual`](https://docs.python.org/3.9/library/unittest.html#unittest.TestCase.assertEqual) which generates nicer error messages on fail.


andyk:
This is so cool! Thanks for figuring this out :D

andyk (commented):
LGTM. Surprisingly straightforward!

PR #234, opened 2022-02-04
Fixes the comments missed in #227 (except for broken Acme agent)

andyk:
I missed 15 code review comments from @nickjalbert in #227. So this PR fixes them.

PR #236, opened 2022-02-05
Fix broken model saving/loading via mlflow in acme agents (fix #233)

### LINKED ISSUE #233: Unexpected run structure when running Acme DQN agent
Copied from PR 227 Review (in a comment hidden by the dumb GH comment accordion), confirmed this happens to me in master:

I'm running into some sort of issue with acme_dqn agent (probably both the acme agents?).  Repro:

```bash
cd example_agents/acme_dqn
rm -rf mlruns/
agentos run agent --param-file parameters.yaml --entry-point learn
```

I get the following (unexpected) run structure:

Screenshot 2022-02-03 134247

In the learning child run, I see metrics and parameters but no artifacts.  In the free floating run, I see the model artifact but no metrics and parameters.  Probably something with the way we terminate and clean up the run?

Additionally, when I then run:

```bash
agentos run agent --param-file parameters.yaml --entry-point evaluate
```

It says it cannot find the saved model from training.

andyk:
This Fixes #233. Update the dqn (and r2d2) agent to successfully load and save to tensorflow, using MLflow as the mechanism for communication between runs.

This is based onto top of #234

### DIFF
--- a/agentos/agent_run.py
+++ b/agentos/agent_run.py
@@ -194,7 +194,7 @@ class AgentRun(Run):
     def end(
         self,
         status: str = RunStatus.to_string(RunStatus.FINISHED),
-        print_results: bool = False,
+        print_results: bool = True,
     ) -> None:
         super().end(status)
         self.log_run_metrics()
--- a/example_agents/acme_r2d2/network.py
+++ b/example_agents/acme_r2d2/network.py
@@ -6,11 +6,13 @@ class R2D2Network:
     def __init__(self):
         self.rnn = BasicRNN(self.environment)
 
-    def restore(self):
-        self.AcmeRun.restore_tensorflow("rnn", self.rnn)
+    def save(self, run=None):
+        # This attribute is set up by AgentOS
+        self.TFModelSaver.save("rnn", self.rnn, run=run)
 
-    def save(self):
-        self.AcmeRun.save_tensorflow("rnn", self.rnn)
+    def restore(self):
+        # This attribute is set up by AgentOS
+        self.TFModelSaver.restore("rnn", self.rnn)
 
 
 # BasicRNN, taken from r2d2 test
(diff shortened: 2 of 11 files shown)


andyk:
@nickjalbert I think this is ready for you to check out. Tests are passing and for me, on Linux, the acme agents seem to be saving and loading checkpoints to/from mlflow as expected.

agentos/agent_run.py
```diff
@@ -192,11 +192,14 @@ def add_episode_data(self, steps: int, reward: float):
         )
 
     def end(
-        self, status: str = RunStatus.to_string(RunStatus.FINISHED)
+        self,
+        status: str = RunStatus.to_string(RunStatus.FINISHED),
+        print_results: bool = False,
```
nickjalbert, line 197: Maybe this should default to `True` so we still get the little console printout when we run the acme agents?

example_agents/acme_r2d2/network.py
```diff
@@ -6,11 +6,13 @@ class R2D2Network:
     def __init__(self):
         self.rnn = BasicRNN(self.environment)
 
-    def restore(self):
-        self.AcmeRun.restore_tensorflow("rnn", self.rnn)
+    def save(self, run=None):
```
nickjalbert, line 9: This appears to not-quite-work for acme_r2d2 because you need to pass the run when you call `save()` (in agent.py and trainer.py) for the model to actually be logged in the run so it can be restored in a later run.  Seems to work once I pass the run!

nickjalbert (changes requested):
LGTM once you pass the run into the save method for acme_r2d2 (see comment); feel free to merge after!

Also, the diff is still weirdly broken on github (showing things already in master).  Hopefully this doesn't mess up the review comments.  Everything looks okay in my local repo when I diff andyk/fix_233 with master, so I suspect the merge will be fine.

example_agents/acme_r2d2/network.py
andyk: Fixed.

agentos/agent_run.py
andyk: Done.

PR #252, opened 2022-02-16
Doc revision: intro and quickstart

nickjalbert:
I decided to get this docs revision party started, so I took a pass over the intro and the quickstart.  I think we should extend the quickstart example to include some pushing and pulling of components; any thoughts on what that script should look like?

Feel free to request change from me OR to just push commits directly to this PR.

As a refresher, to build and view the docs:

```bash
python scripts/build_docs.py
open agentos/docs/0.1.0-alpha/index.html
```


andyk:
AWESOME! Shipped it :D

PR #261, opened 2022-02-24
Small developer experience improvements

andyk:
* add Repo.from_github()
* use content hash in name of auto-generated source files from py shell
* add more core objects to the root agentos namespace to be easily accessible


agentos/repo.py
```diff
@@ -62,6 +62,21 @@ def from_spec(spec: NestedRepoSpec, base_dir: Path = None) -> "Repo":
...
+    ) -> "Repo":
+        return cls.from_spec(registry.get_repo_spec(identifier))
+
+    @classmethod
+    def from_github(
+        cls, github_account: str, repo_name: str, identifier: str = None
+    ) -> "GitHubRepo":
+        if not identifier:
+            identifier = f"{github_account}/{repo_name}"
+        url = f"https://github.com/{identifier}"
```
nickjalbert, line 77: If I pass `identifier` here, won't this result in a broken github url?  E.g. I think the call:

```python
 Repo.from_github('agentos-project', 'agentos', identifier='foo')
```

result in a broken repo object.

nickjalbert (approved):
LGTM other than my one comment.  Feel free to merge after fix!

agentos/repo.py
andyk: Fixed!

PR #263, opened 2022-02-25
Rename ParameterSet to ArgumentSet

### LINKED ISSUE #262: Rename ParameterSet to ArgumentSet
Programming languages tend to call them args when the are being passed in at function call time, and parameters when you are defining the function. Our use case for what we currently call a `ParameterSet` is the former so renaming them to `ArgumentSet` makes sense. Also, parameters are more overloaded in ML, since model parameters (e.g. mlflow.log_parameter) is a different concept than a run argument.

andyk:
Fixes #262


nickjalbert (approved):
LGTM!  There are a few places where we e.g. still refer to `params` in the code (instead of the potentially-less-confusing  `arg_set`).  I think it'll be quick for me to go through and update so I'll merge and then generate a new PR that does that.

PR #264, opened 2022-02-25
Make it easier (and more intuitive?) to use `Component.run()`

andyk:
NOTE: this is based on #263. Let's merge that first!

We've been talking about making`Component.run()` easier to use by getting rid of the burden of building an `ArgumentSet`. This does that by renaming the existing version of the function (which has lots of knobs and requires an ArgumentSet) to be `Component.run_with_arg_set()` and having the `run()` function itself be simpler and behave more like the entry point function itself, e.g.: you just pass the entry point args directly into the `run()` function and get back the return value.

The trade-off is that you don't get a pointer back to the ComponentRun that was created.

This also adds a convenience function `Registry.from_repo()`.

Demo:

```python
import agentos
repo = agentos.Repo.from_github("agentos-project", "agentos")
reg = agentos.Registry.from_repo(repo, "example_agents/random/components.yaml", "master")
comp = agentos.Component.from_registry(reg, "agent")
comp.run("run_episodes", num_episodes=10000)
```

### DIFF
--- a/agentos/cli.py
+++ b/agentos/cli.py
@@ -175,7 +175,7 @@ def run(
         arg_set = ArgumentSet.from_yaml(param_file)
         entry_point = entry_point or component.get_default_entry_point()
         arg_set.update(component_name, entry_point, param_dict)
-        run = component.run(entry_point, arg_set)
+        run = component.run_with_arg_set(entry_point, arg_set)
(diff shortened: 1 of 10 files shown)


nickjalbert (approved):
Cool!  That does feel cleaner when I run through the demo.  

I had to resolve a conflict in master, so I'll merge once the build is green.

PR #290, opened 2022-03-02
WIP: Rework component identifier

### LINKED ISSUE #243: Replace `Component.Identifier` uses with `ComponentIdentifier`
I know I suggested we use an inner class for `ComponentIdentifier` in the first place, but now that we've been working with it for a while, I feel like it's confusing to have both `ComponentIdentier` and `Component.Identifier` and we should delete the inner class `Component.Identifier` (which is just an alias anyway) and replace all uses of it with `ComponentIdentifier`.

andyk:
* Delete Component.Identifier and replace all uses of it with `ComponentIdentifier` -- fixes #243 
* Simplify `ComponentIdentifier` by making it a subtype of `str`


andyk:
@nickjalbert let me know what you think of this approach?

nickjalbert:
@andyk I still have a vague feeling of uneasiness around subclassing `str`, but we've basically gotten most of the way there already and I haven't been stricken by any unexpected bugs using `ComponentIdentifier` so my feeling is probably unjustified.  I like that this eliminates some code, too.  

So, looks like a good approach to me!  Let's be on the look out for any surprising gotchas, though.

Issue #280, opened 2022-03-01
Integrate VirtualEnv Management into Component


nickjalbert:
Right now, the virtualenv management is completely separated from the Component lifecycle.  For example, in [the CLI](https://github.com/agentos-project/agentos/blob/0f3d437f9ca11b5f1d01308b8f81de578bb18b4c/agentos/cli.py#L170), we:

1. Get the virtualenv
2. Activate the virtualenv
3. Get the Component
4. Run the Component's entry point

However, there's duplicate work going on here because to get the virtualenv (step 1), we need to construct the Component DAG to find all the requirements files associated with the Components.  That DAG then gets tossed and then **reconstructed** in step 3 to get Component.

This implementation was chosen due to some awkwardness around Component instantiation: you need the virtualenv before you import the managed class, and currently Component eagerly imports the managed class in a way that makes it difficult to fully construct the virtualenv beforehand.

However, with the changes proposed in #276, virtualenv management becomes much cleaner as something that happens completely within the Component.  This also makes REPL based demos cleaner.

nickjalbert:
@andyk FYI with the merge of #293, I'm going to work on this.

PR #293, opened 2022-03-02
Import and instantiate Components in get_object()

### LINKED ISSUE #276: Component class should create managed instance JIT instead of at __init__ time.
Currently, `Component.__init__()` accepts as required argument an instance of a class to be managed. This made more sense when `Component.get_object()` always instantiated the class. However, now we have generalized the Component abstraction from being a "managed class instance" to being a "managed object" which could be a class or class instance (and soon we will modules also).

So now, it would make more sense to have the Component class create the managed object just in time, i.e. as part of the call to `get_object()`.

nickjalbert:
Fixes #276; moves Component import and instantiation in `get_object()`.  Clears the way for work on #280 .

### DIFF
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -43,7 +43,9 @@ def test_component_repl_demo():
         GenericDependency, instantiate=False
     )
     class_comp_with_diff_name = Component.from_class(
-        GenericDependency, identifier="ClassDependency", instantiate=False,
+        GenericDependency,
+        identifier="ClassDependency",
+        instantiate=False,
     )
 
     # Add dependencies to SimpleAgent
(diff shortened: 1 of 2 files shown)


tests/test_component.py
```diff
@@ -55,7 +57,7 @@ def test_component_repl_demo():
 
     assert "GenericDependency" in agent_comp.dependencies.keys()
     inst_dep_obj = agent_comp.dependencies["GenericDependency"].get_object()
-    assert type(inst_dep_obj) == GenericDependency
+    assert inst_dep_obj.__class__.__name__ == "GenericDependency"
```
andyk, line 60: Aren't these equivalent? Why the switch to this alternate syntax? Or maybe I'm missing something.

andyk (approved):
Great refactor, super clean! :D

PR #294, opened 2022-03-03
fix bug in build_docs

andyk:
Fix a small bug I found in the build_docs script when releasing 0.1.0

### DIFF
--- a/scripts/build_docs.py
+++ b/scripts/build_docs.py
@@ -64,7 +64,6 @@ if known_args.watch:
     build_tool = "sphinx-autobuild"
 else:
     build_tool = "sphinx-build"
-print(os.getcwd())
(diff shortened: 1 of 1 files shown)


scripts/build_docs.py
```diff
@@ -64,10 +64,13 @@ def update_latest_symlink():
     build_tool = "sphinx-autobuild"
 else:
     build_tool = "sphinx-build"
-build_cmd = [build_tool, docs_dir, versioned_build_dir, "-c", "documentation"]
+print(os.getcwd())
```
nickjalbert, line 67: We probably don't need this print statement.

nickjalbert (approved):
LGTM!  My only thought you might remove the `print(os.getcwd())` statement.

PR #295, opened 2022-03-03
Add "module Components", auto-inferred registry from repo, and add default version to Repo.

### LINKED ISSUE #267: Git[Hub]Repos should have a default version (i.e., branch)
The same branch that is used when you git clone the repo (probably `master`, or I think in some more recent repos, `main`).

This was inspired by my experience having to set the branch in my demo in #264, where we have to pass in a version to most of `Repo`'s functions.  

Here is an example:

```python
>>> import agentos
>>> sb3_repo = agentos.Repo.from_github("DLR-RM", "stable-baselines3")
>>> sb3_repo.get_local_repo_dir()
get_local_repo_dir() missing 1 required positional argument: 'version'
```

I would expect `sb3_repo.get_local_repo_dir()` to just assume `master` or `main` unless I [optionally] pass a version.

### LINKED ISSUE #278: Allow a Component to be a managed class instance, managed class, or managed module
Currently, a Component can be either a managed class or a managed class instance. The `instantiate: bool` decides which it is, per:

https://github.com/agentos-project/agentos/blob/0f3d437f9ca11b5f1d01308b8f81de578bb18b4c/agentos/component.py#L40-L48

Allowing a component to manage a module would require relatively small change (we already load the module in) and unlock a significant functionality & flexibility.

I propose we support this by making the `class_name` parameter to `Component.__init__()` (and any related factory methods) optional, and then if `class_name` isn't provided then the component is a managed module.

### LINKED ISSUE #304: Generate registry by inspecting a repo
This issue depends on #277 and #278.

To make creation of Components much easier, we can have a function `Registry.from_repo_inferred()` that scans the files in a repo and generates an `InMemoryRegistry` that contains a Module Component (requires #278) for each of the python files in the repo. We could also generate a component for each of the classes (in all of the modules) in the Repo as well as a Component for an instance of each of those classes.

This function can also look to see if a "requirements.txt" or "setup.py" file exists and use those to ensure availability of dependencies. Note that it should create a new PyPIRepo (see #303) in the generated registry for each of the dependencies.

Then we can have the higher level `Registry.from_repo(registry_file="components.yaml")` function:
1. if `registry_file` exists in the repo, load that file as an InMemoryRegistry and return it
2. else, call `Registry.from_repo_inferred` to create a registry based on inspection.

This will make demos much more concise and compelling as well.

### LINKED ISSUE #307: `Component.to_spec()` not serializing `instantiate` field to spec
The spec emitted by `Component.to_spec()` does not include `instantiate`, but should.

andyk:
* Fixes #278
* Fixes #267
* Fixes #304 
* Fixes #307
* Fixes a few docstring typos I noticed while reviewing #306

The changes in this PR allow a Component to be a managed module, as an alternative to the existing support for a Component being a managed class or class instance.

This also adds reworks `Registry.from_repo()` which now wraps conditional logic that in turn either calls `Registry.from_file_in_repo()` (which is where the code that was previously in `Registry.from_repo()` moved to) or `Registry.from_repo_inferred()` which is a new function that recursively searches through all files in a repo looking for all python files and then creating module components out of them in a new registry.

To see how both of these new features working together, try:
```python
>>> import agentos
>>> repo = agentos.Repo.from_github("agentos-project", "agentos")
>>> reg = agentos.Registry.from_repo_inferred(repo, requirements_file="dev-requirements.txt")
>>> assert "module:agentos__component.py" in reg.to_dict()["components"]
```
For a more extensive example, check out the `test_module_component_from_agentos_github_repo()` in `test_component.py`.


andyk:
@nickjalbert this is now ready for a review!

agentos/registry.py
```diff
@@ -83,6 +84,54 @@ def from_repo(
         ), "YAML is the only registry file format supported currently"
         return cls.from_yaml(repo.get_local_repo_dir(version) / file_path)
 
+    @classmethod
+    def from_repo_inferred(
+        cls,
+        repo: "Repo",
+        py_file_suffixes: Sequence[str] = [".py", ".python"],
```
nickjalbert, line 91: It might be good to [avoid mutables-as-default-args](https://stackoverflow.com/questions/9158294/good-uses-for-mutable-function-argument-default-values) unless we're taking advantage of their somewhat-surprising behavior. 

Maybe default this to `None` and then assign it this list in the body if it is, in fact, `None`.

agentos/registry.py
```diff
@@ -83,6 +84,54 @@ def from_repo(
...
+        requirements_file: str = "requirements.txt",
+    ):
+        from agentos.component import Component  # Avoid circular ref.
+
+        reg = InMemoryRegistry()
+        # get list of python files in Repo
+        py_files = set()
+        for suff in py_file_suffixes:
+            found = repo.get_local_repo_dir().rglob(f"*{suff}")
```
nickjalbert, line 100: ~~Does rglob recursively search in directories?~~  Answered my own question by running through the demo.

agentos/repo.py
```diff
@@ -306,13 +319,15 @@ def to_spec(self, flatten: bool = False) -> Dict:
         }
         return flatten_spec(spec) if flatten else spec
 
-    def get_local_repo_dir(self, version: str) -> Path:
+    def get_local_repo_dir(self, version: str = None) -> Path:
+        version = version if version else self._default_version
```
nickjalbert, line 323: Style note: I think you're doing this the canonical Python way, but I've been moving to `version = version or self._default_version` when I'm fine with getting `self._default_version` for any False-y value of `version`.

I think either way is fine (let me know if you have an opinion here), but the `or` way might save a little typing (and sometimes I find these ternary expressions a little harder to read).

nickjalbert (approved):
Looks good to me!!  The inference is pretty cool. :)

Left a few comments; nothing big.  Feel free to merge whenever.

agentos/component.py
```diff
@@ -38,26 +38,28 @@ def __init__(
         self,
         repo: Repo,
         identifier: ComponentIdentifier,
-        class_name: str,
         file_path: str,
+        class_name: str = None,
+        instantiate: bool = None,
```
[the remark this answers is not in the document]
andyk: I agree it is a bit confusing, so I just changed the default value of `instantiate` to `False`.

A little more background/context on my thinking here:

I though about that and originally hesitated to change the default setting from `True` to `False` since that meant updating all of our example/test registries/specs that omit `instantiate` from their spec (assuming omission implies `True`)... because I was aiming to keep the current default behavior _when `class_name` is provided_.

So then I thought about making the default `True`. However, now that `class_name` is optional, I don't think it makes sense for `instantiate` to default to `True` when `class_name` is `None` (i.e., for a Module Component).

That's what had led me to making `instantiate`'s default value conditional on class_name, which is the implementation you commented on.

After reading your comment though, I decided to just bite the bullet and make the default value `False` which officially makes Components more of a managed Module by default, with the option to add a `class_name` attribute to make the Component a managed Class, and the option to _**additionally**_  set `instantiate` to `True` to make the Component a managed Class Instance.

I've also updated the Component docstring summary of what a Component is and also the param descriptions.

agentos/registry.py
andyk: Good catch. I changed it to:
```python
        py_file_suffixes: Tuple[str] = (".py", ".python"),
```
since tuple is immutable.

agentos/repo.py
andyk: Hmm, I don't feel strongly about it but I actually find it takes more brain power to think about the semantics of the `or` operator than to read the inline conditional assignment statements.

It's a bit non-intuitive to me that `"hi" or "bye"` results in `"hi"` rather than `True`. But that's probably because I'm just not used to this usage of `or` in Python, and haven't read much about how `or` works under the hood in Python.

agentos/component.py
```diff
@@ -196,18 +204,18 @@ def from_registry_file(
     @classmethod
     def from_class(
         cls,
-        managed_cls: Type[T],
+        managed_obj: Type[T],
```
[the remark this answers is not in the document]
andyk: Good observation. I renamed the parameter to `class_obj` since `managed_cls` might make it seem like that specific Class object might become managed, which is how it used to work. Now the Component creates a new Class object for every call to `run()` or `get_object()`.

andyk:
Ok, I think I've addressed all of your feedback @nickjalbert. Since the change to the default value of the `instantiate` parameter (from `True` to `False`) to `Component.__init__()` was a pretty big change, I want you to sign-off before merge.

agentos/component.py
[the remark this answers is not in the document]
nickjalbert: I like this choice; I think it makes sense in terms of Components as more general object managers (even though it does require a bit of re-wiring of what we think of as the default).

agentos/registry.py
nickjalbert: 👍

agentos/repo.py
nickjalbert: Yeah seems mostly like a matter of taste and what you're used to.  

Tidbit: I ran across a wiki article on this type of operator ([Null coalescing operator](https://en.wikipedia.org/wiki/Null_coalescing_operator)) and it notes that Python's `or` isn't a true implementation because, as mentioned above, you get the right hand side of the expression if the left side is any sort of False-y value.

nickjalbert:
LGTM!  I'm going to merge this. :)

PR #302, opened 2022-03-05
Add isort as part of code formatting

nickjalbert:
I was reading some discussion about Python black and people mentioned [isort](https://pypi.org/project/isort/) as a nice complement, so I added it to our formatting script.

Isort auto-sorts import statements alphabetically while also dividing them up as recommended by [PEP8](https://www.python.org/dev/peps/pep-0008/#imports).  

You will probably have to run `pip install -r dev-requirements.txt` in your dev env again to install isort so `python scripts/format_code.py` works as expected.

Let me know what you think!


andyk (approved):
Cool. I like it

PR #313, opened 2022-03-11
RunCommand now tracks log_return_value

andyk:
Also, add `LOG_RETURN_VALUE` to `RunCommandSpecKeys`
This also renames `RunCommandSpecKeys.PARAMETER_SET` to `ARGUMENT_SET` which we missed in past renaming effort.

### DIFF
--- a/agentos/run_command.py
+++ b/agentos/run_command.py
@@ -44,10 +44,12 @@ class RunCommand:
         component: "Component",
         entry_point: str,
         argument_set: "ArgumentSet",
+        log_return_value: bool,
     ):
         self._component = component
         self._entry_point = entry_point
         self._argument_set = argument_set
+        self._log_return_value = log_return_value
 
     def __repr__(self) -> str:
         return f"<agentos.run_command.RunCommand: {self}>"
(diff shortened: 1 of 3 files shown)


andyk:
@nickjalbert note that this adds a database migration to the webapp it'll require a push to Heroku. BTW, I was playing with to_registry() and accidentally pushed a Repo spec to the Heroku app, and was delighted to see it worked :)

andyk:
@nickjalbert This is ready for a review! This PR includes some bug fixes and small features that I've been encountering while working on the demo.

I've added my current demo script to #283 -- and it currently works for:
1. inferring a registry from SB3
2. creating a agent component from that registry
3. run that agent using agentos
4. use the run_command generated from the initial run to do a re-run.
5. publish the run to different types of registries (local file and WebRegistry)
6. read in the run from those different registries, and re-run it

agentos/run_command.py
```diff
@@ -44,10 +44,12 @@ def __init__(
         component: "Component",
         entry_point: str,
         argument_set: "ArgumentSet",
+        log_return_value: bool,
     ):
```
nickjalbert, line 48: Would it make sense to add a docstring here?  I guess `log_return_value` is pretty self-explanatory, but maybe it's not super clear why you wouldn't always set this to `True`.

andyk: Added a docstring to `__init__()`

PR #323, opened 2022-03-12
Scripts for our code demos to users

andyk:
Do not merge me.

@nickjalbert -- I thought it might be nice to keep these somewhere we can collaborate on them via version control, just not sure where that is :D

Perhaps these scripts should be incorporated into the docs somehow in a way where they are still runnable?

Probably don't want to make tests out of them since currently the inference takes order 5min.


nickjalbert:
> I thought it might be nice to keep the demo scripts somewhere we can collaborate on them via version control, just not sure where that is

SGTM!  Maybe we could put them under an `agentos/demo/` folder for now?

> Perhaps these scripts should be incorporated into the docs somehow in a way where they are still runnable?

At least, they seem like they could serve as the foundation for a revamped quickstart or a tutorial.  Not sure if there's a great way to automatically incorporate them, but let's think about that a bit more...

nickjalbert (approved):
LGTM!  Not quite sure why code formatting is broken, but feel free to merge after everything is green.

PR #325, opened 2022-03-15
Only activate a Component's venv once

### LINKED ISSUE #324: VirtualEnvs should preserve sys.path changes
Sometimes when imported, a library will make changes to the `sys.path` upon which other code in that library depends (I believe `ray`  or one of its dependencies does this).  

Right now we use an idiom like the following for importing and running managed objects:

```python
with venv:
    import_managed_obj()
```

This can cause errors because when we exit the venv context, we lose all changes that the managed object has made to the environment.  Instead, we should preserve changes to the environment.

Additionally, I think we want to generally encourage Components to activate the shared virtualenv once (and not deactivate it) due to potentially unexpected behavior around PCS not being able to provide full environment isolation.  I'm not sure if this once-only behavior should be explicitly enforced.

### Related: Why PCS can't provide fully environment isolation

The [Just-in-time virtual env](https://docs.google.com/document/d/1i1AoTF63IUh9hgYBgxdBRsbBgwD8YvKJ9HbwRwR_L68/edit#heading=h.5skdbrwtmflo) design doc goes more in-depth into some of the limitations of Python virtual env manipulation.  From the doc:

>* Python has facilities for extension modules, these are compiled C and C++ libraries (.so files on linux systems) that you can import into your Python code in the same way you would import a pure Python module.
> * These extension modules, however, are treated differently than pure Python modules.  Many of these modules use shared process-wide state (instead of module-local state like pure Python modules).
> * This shared process-wide state prevents us from cleanly reloading these modules when, for example, we want to import a different version of an already imported extension module so that a Component can run against this new version.
> * Currently, there are no facilities for reloading extension modules although there is a proposal (and maybe active work) on making extension modules default to a per-module state model instead of a per-process state model.2  This would potentially make reloading more feasible but presumably all extension modules that use the current per-process style would have to be rewritten.
> * As a result, we are only able to isolate a Component’s environment from other Component environments when all modules are pure Python (i.e. no extension modules are used)
> * Unfortunately, many useful real-world libraries require extension modules (e.g. protobuf which is used in MLflow, tensorflow).
> * Trying our environment isolation techniques when extension modules are loaded result in unexpected errors (e.g. classes that should be the same are not the same, data that should only be initialized once gets re-initialized) and break the running of our Components.

nickjalbert:
This PR makes it so that when we're building and running a Component DAG with one or more `requirements_path`'s defined, we only activate the virtual env built out of those requirements once and never deactivate.  

I think this is the most reasonable behavior given that:

* It is impossible to isolate dependencies for different components in CPython (see write up [here](https://docs.google.com/document/d/1i1AoTF63IUh9hgYBgxdBRsbBgwD8YvKJ9HbwRwR_L68/edit#heading=h.5skdbrwtmflo))
* Some libraries make changes to the environment when they are imported that they expect to persist (e.g. manipulation of the `sys.path`)
* We support the `get_object()` flow and that object expects to run in its environment.

I added a local check to `Component.add_dependency()` to make sure you weren't doing obviously bad things (adding a Component with requirements to a DAG after a venv was created and activated), but I didn't add any global checks.  Specifically, you could create two Components with different venvs and activate them sequentially and potentially get unexpected behavior.  This is a failure mode we might want to address in the future.

Fixes #324

### DIFF
--- a/agentos/component.py
+++ b/agentos/component.py
@@ -92,6 +92,7 @@ class Component:
         self.instantiate = instantiate
         self.dependencies = dependencies if dependencies else {}
         self._use_venv = use_venv
+        self._venv = None
         self._dunder_name = dunder_name or "__component__"
         self._requirements = []
         self._parent_components = set()
@@ -394,13 +395,18 @@ class Component:
         assert fn is not None, f"{instance} has no attr {function_name}"
         fn_args = arg_set.get_function_args(self.name, function_name)
         print(f"Calling {self.name}.{function_name}(**{fn_args})")
-        with self._build_virtual_env():
-            result = fn(**fn_args)
+        result = fn(**fn_args)
         return result
 
     def add_dependency(
         self, component: "Component", attribute_name: str = None
     ) -> None:
+        if self._venv and component.requirements_path:
+            raise Exception(
+                "You cannot add a dependency with a requirements_path after a "
+                "virtual env has already been constructed and activated. Try "
+                "restarting Python and rebuilding your dependency graph"
+            )
         if type(component) is not type(self):
             raise Exception("add_dependency() must be passed a Component")
         if attribute_name is None:
@@ -443,6 +449,9 @@ class Component:
 
     def _import_object(self):
         """Return managed module, or class if ``self.class_name`` is set."""
+        if not self._venv:
+            self._venv = self._build_virtual_env()
+            self._venv.activate()
         full_path = self.repo.get_local_file_path(
             self.file_path, self.identifier.version
         )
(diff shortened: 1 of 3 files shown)


andyk:
LGTM. Feel free to merge!

One observation: since this mostly changes how `Component` uses `VirtualEnv`, it is still possible for a user to directly create and activate two `VirtualEnv`s from within the same python process, and those could try to load c modules that conflict (e.g., two different versions of the same c module), right?

In the future, we may want to change the architecture of `VirtualEnv` to only allow a single instance of `VirtualEnv` within its Python process, not just those created by a `Component`.

andyk:
Also, looks like maybe some conflicts w/ master after I merged #313

andyk (approved):

nickjalbert:
> it is still possible for a user to directly create and activate two VirtualEnvs from within the same python process, and those could try to load c modules that conflict (e.g., two different versions of the same c module), right?

Right.  If the user only used pure Python modules (that is, the module and all its dependencies are written in pure Python), then I believe you could clear the `sys.modules` cache, change the environment, and then import a different version of the module in a different venv.  I'm pretty sure I had an example of this working (maybe for numpy?) when we started looking at trying to isolate multiple virtual envs.

But, because C/C++ modules often used shared state, you can't reload them and so you can get conflicts or unexpected behavior if you try to load modules that require different versions of the same C/C++ module.

> In the future, we may want to change the architecture of VirtualEnv to only allow a single instance of VirtualEnv within its Python process, not just those created by a Component.

Yeah, I thought about this.  I then wondered if there was a use case where you'd want to activate multiple venvs that had non-conflicting requirements.  Not sure if there would be a reason to do that, though.  So, basically, I'm open to doing this but want to think a little more about the implications.

nickjalbert:
@andyk thanks for the review!  Fixed the conflicts; will merge when green.

PR #326, opened 2022-03-19
Fixes broken install of dev-requirements on apple silicon (Apple M1 Macbook Pro)

andyk:
Up until now,  on my Macbook Pro M1 I've been ignoring the rllib tests which have failed with a segfault.

However I recently:
1. switched to the arm64 version of brew (I previously had installed the x86 vesion which I guess was using rosetta for much of what it was installing?)
2. used that new brew to re-install conda
3. made a new env
4. installed scipy via conda (since installing scipy via pip tries compiling numpy on apple silicon which fails)
5. tried pip installing sb3 requirements, and this fails because apparently torch 1.9.1 is not available on MacOS w/ apple silicon, but 1.9.0 is.

so this changes the version of torch in the example_agent requirements.txt file to 1.9.0, which fixes the error on my Macbook Pro (MBP) M1.

Also, this PR updates the string we look for in sys.platform from "windows" to "win32". I looked at the test output on the github workflow windows machine and it was showing:

```
Ignoring torch: markers 'sys_platform == "linux"' don't match your environment
Ignoring torch: markers 'sys_platform == "darwin"' don't match your environment
Ignoring torch: markers 'sys_platform == "windows"' don't match your environment
```

Which seemed odd since I would expect it to be matching "windows", so I looked it up and per https://docs.python.org/3/library/sys.html?highlight=sys#sys.platform "win32" is the correct string, not "windows".

So this PR also fixes our use of that string in two example_agent requirements.txt files.

### DIFF
--- a/example_agents/rllib_agent/requirements.txt
+++ b/example_agents/rllib_agent/requirements.txt
@@ -3,7 +3,7 @@
 # Pass the following option to pip on linux to get the CPU only torch
 # -f https://download.pytorch.org/whl/cpu/torch_stable.html
 torch==1.10.2+cpu; sys_platform == 'linux'
-torch==1.10.2; sys_platform == 'windows'
+torch==1.10.2; sys_platform == 'win32'
 torch==1.10.2; sys_platform == 'darwin'
 ray[rllib]==1.10.0
 gym==0.21.0
--- a/example_agents/rllib_agent/requirements.txt
+++ b/example_agents/rllib_agent/requirements.txt
@@ -6,7 +6,7 @@ torch==1.10.2+cpu; sys_platform == 'linux'
 torch==1.10.2; sys_platform == 'win32'
 torch==1.10.2; sys_platform == 'darwin'
 ray[rllib]==1.10.0
-gym==0.21.0
+gym==0.19.0
 
 # agentos_pip_cmdline:
 #   linux:
--- a/example_agents/sb3_agent/requirements.txt
+++ b/example_agents/sb3_agent/requirements.txt
@@ -7,4 +7,4 @@ dm-env==1.5
 mlflow==1.23.1
 torch==1.9.1+cpu; sys_platform == 'linux'
 torch==1.9.1; sys_platform == 'windows'
-torch==1.9.1; sys_platform == 'darwin'
+torch==1.9.0; sys_platform == 'darwin'
--- a/example_agents/sb3_agent/requirements.txt
+++ b/example_agents/sb3_agent/requirements.txt
@@ -6,5 +6,5 @@ stable_baselines3==1.4.0
 dm-env==1.5
 mlflow==1.23.1
 torch==1.9.1+cpu; sys_platform == 'linux'
-torch==1.9.1; sys_platform == 'windows'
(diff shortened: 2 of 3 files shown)


andyk:
Also, interestingly, after installing scipy via conda, and then running `python install_requirements.py` on my M1 (and having it work in this branch), I get the following error when I try to import ray in my python shell:

```
>>> import ray
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/ray/__init__.py", line 36, in _configure_system
    import grpc  # noqa: F401
  File "/opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/grpc/__init__.py", line 22, in <module>
    from grpc import _compression
  File "/opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/grpc/_compression.py", line 15, in <module>
    from grpc._cython import cygrpc
ImportError: dlopen(/opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/grpc/_cython/cygrpc.cpython-39-darwin.so, 2): Symbol not found: _CFRelease
  Referenced from: /opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/grpc/_cython/cygrpc.cpython-39-darwin.so
  Expected in: flat namespace
 in /opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/grpc/_cython/cygrpc.cpython-39-darwin.so

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/ray/__init__.py", line 83, in <module>
    _configure_system()
  File "/opt/homebrew/Caskroom/miniforge/base/envs/agentos_dev2/lib/python3.9/site-packages/ray/__init__.py", line 38, in _configure_system
    raise ImportError(
ImportError: Failed to import grpc on Apple Silicon. On Apple Silicon machines, try `pip uninstall grpcio; conda install grpcio`. Check out https://docs.ray.io/en/master/installation.html#apple-silicon-support for more details.
```

nickjalbert (approved):
LGTM!  Thanks for fixing those string.   

As discussed, before merge maybe we unify the sb3 torch requirements to also use 1.10.2 because that seems to work and has an M1 wheel?

> `pip uninstall grpcio; conda install grpci`

Also, how hard would it be to make `install_requirements.py` script automatically do this uninstall/reinstall on M1s?  Or does that assume too much about the environment (e.g. conda existing).

andyk:
I changed updated the sb3 torch requirement to 1.10.2 and updated `install_requirements.txt` to handle the case of apple-silicon.

I also noticed a conflict with the gym requirements between sb3 (which wants <0.20) and rllib_agent (which wants 0.21.0) so I changed rllib_agent to use 0.19.0 and tested that it didn't break anything on my Macbook.

PR #328, opened 2022-03-19
WIP: Break pcs out into separate dir (i.e., module)

andyk:


andyk:
@nickjalbert I think this and #326 are ready for review!

example_agents/rl_agents/README.txt
```diff
@@ -0,0 +1,17 @@
+# Installing tensorflow on Apple Silicon
```
nickjalbert, line 1: Maybe add a brief description here on how to run the agents in this directory.

nickjalbert (approved):
LGTM except it appears `python scripts/build_docs.py` is broken (it's still trying to import `agentos/version.py` instead of `pcs/version.py`).  We'll also have to update the autodoc stubs in `documentation/api` to point to the new pcs location.

Not sure if that's what you meant by the docs not being updated, but we could port that in a different PR if you want to merge this one now.

example_agents/rl_agents/README.txt
andyk: done.

andyk:
Ok @nickjalbert I think I've addressed all of your feedback and updated the main text of this PR to reflect the updates. If tests are green I think we are in good shape, though might be nice if you try out the newly updated `scripts/build_docs.py` file.

andyk:
Actually I just noticed this PR still doesn't add readme files to all of the example agents, which is one of the things you suggested.

nickjalbert:
Build is green!  Merging!  Thanks for doing this. :)

PR #336, opened 2022-03-22
WIP: Spec structure revamp

### LINKED ISSUE #250: Simplify & unify what a "Spec" is
Summary:
* Create common `SpecObject` base type that `Repo`, `Component`, `RunCommand`, `Run`, `ComponentRun`, and `AgentRun` extend.
* `Registry` should
    * have functions that handle (e.g., add/get) specs generically (i.e., all types of specs)
    * support `identifier in my_registry` syntax
* `SpecObject` should
    * define `to_spec()`, `from_spec()`, `to_registry()` (including the hairy recurse/force logic that is currently duplicated lots of times), `from_registry()`, and `publish()`
    * handle dependencies in a consistent way. E.g., a Component can optionally depend on other Components and must depend on a Repo, and these dependencies should be handled via the same mechanism.
    * `define `identifier` as hash of spec contents
* it is unclear if `SpecObject` should support versioned identifiers or versioning vs that being handled at the Component Repo semantic layer.

Currently the Registry API contains functions that are specific to the different "types" of specs we support.

Specs can currently be of type "repo", "component", "run_command", "run".  Also repo specs have a 'type' field which can be either "local" or "github".

This means that new subclasses of Registry, e.g. WebRegistry, have to implement functionality for each of the different types of specs. For example: `WebRegistry.add_repo_spec()`, `WebRegistry.add_component_spec()`, etc., and same for the `get_x_spec()` family of functions and also the `contains_x_spec()` functions.

But since specs all are very similar in nature (essentially a dictionary where keys are always strings and values are basic/simple things that are easy to serialize into YAML like strings & numerics), these functions all end up being similar across the different spec types.

It might simplify things conceptually, and allow us to clean up a lot of similar looking code, if we unified all specs and instead of having a Registry keep a different dictionary for each of the types, a Registry could have just one dictionary mapping identifier to spec_body (which is a dict). And in the spec body, we could require the key 'type' with a str value in ["local_repo", github_repo", "component", "run_command", "run", "component_run", "agent_run"].

Then we could have `Registry.add_spec()` and `Registry.get_spec(identifier)` handle all spec types. For convenience we could decide to keep type-specific functions like `Registry.add_repo_spec()` that wrap the more generic `add_spec()` or `get_spec()` functions.

For human readability, we could also optionally allow Registry yaml files to group the specs by type.

This would also allow us to support python's "x in y" syntax for Registry objects, e.g., `[arbitrary identifier of any spec type] in my_registry`.

This would also make it easier to add new spec types.

Finally, we should consider whether `.to_spec()` can infer the fields to include in the resulting spec automatically from the corresponding Python class in the AgentOS project, e.g.,  can `agentos.run.Run.to_spec()`, which returns a `RunSpec`, be a generic function that the `Run` inherits?

More specifically, `.to_spec()` could be the same for all spec types (e.g., Repo, Component, etc.) and based off of the either conventions related to the structure of the class or an explicit decorator that a developer uses to mark an attribute as a `@spec_field`. We could achieve this by having a `SpecObject` base class where `.to_spec()` and `to_registry()` are defined so that Repo, Component, RunCommand, and Run could inherit them by subclassing `SpecObject`. Thus we could avoid each having to implement their own version of `to_registry()` and `to_spec()`.

andyk:
* Fixes #250.
* Significant rename of core PCS and AgentOS abstractions (classes and file names)


andyk:
@nickjalbert, as we discussed on the AgentOS discord channel, I'm switching this PR from a draft to an actual PR so that we can get it merged and start finishing up the bugfixes as new PRs against master.

The acme ports (and getting acme tests working) are still a work in progress. Before we merge this, I think I should comment out the acme and web tests, and plan to fix those in follow-on PRs, and we should try to get all other tests to pass for a green build.

What do you think?

PR #356, opened 2022-04-11
Run table in webapp

### LINKED ISSUE #333: Webapp homepage should show table of AgentRuns when Component is an Env or Agent

### LINKED ISSUE #353: Add demo that runs an example agent & publishes the resulting AgentRun to a WebRegistry server running locally
Currently, running a example agent using the included registry file (`components.yaml`) creates a Run of a Component with a LocalRepo, which can't be "published" (i.e., added to a WebRegistry) because it doesn't provide reproducibility. Specifically, the `WebRegistry.add_repo_spec()` requires the spec that is passed in to have a `url` attribute.

`Component.to_frozen_registry()` handles this for the case of a component that has a dependency on a local repo, so we should be able to use this or extend it as necessary in a demo.

andyk:
Fixes #333, fixes #353

* Adds a demo that creates and runs the example sb3 agent and then publishes the associated ComponentRun to the WebRegistry.
* Updates webapp homepage to show a list of environments that have been included in published AgentRuns, and for each, the metrics from every Run.
* Updates documentation to show this new demo and also fix a few broken references to code in documenation/docstrings.


andyk:
@nickjalbert i think this is ready for a review!

documentation/demos/index.rst
```diff
@@ -11,10 +11,24 @@ AgentOS WebRegistry django server app, which you can start via::
...
 
-AgentOS demo agent from Github
-==============================
-In this demo, we compose an agent out of components that we fetch from
-the AgentOS github repo, then we run that agent and publish the run to
+AgentOS example SB3 agent from registry file in Github
+======================================================
+In this demo, we create an agent Component from a registry file that
+exists in our local clone of the AgentOS github repo. Then we auto-convert
+the agent Component from an unversioned component to a versioned component
```
nickjalbert, line 18: "." after "component"?

documentation/overview.rst
```diff
@@ -1,5 +1,5 @@
 ***********************************
-Component System Overview
+Python Component System Intro
```
nickjalbert, line 2: Maybe we should change this title to something like "Python Component System (PCS) Intro"?  Otherwise we don't explicitly define the PCS acronym on this page before using it.

example_agents/acme_dqn/agent.py
```diff
@@ -20,8 +20,12 @@ def __init__(self, **kwargs):
         )
 
     def evaluate(self, num_episodes):
-
-        with self.AcmeRun("evaluate", active_component_run(self)) as run:
+        with self.AcmeRun(
+            run_type="evaluate",
```
nickjalbert, line 24: This was my bad I think, but could you swap out these magic `run_type` strings in the example agents for the constants in the run class?

pcs/component.py
```diff
@@ -39,9 +39,10 @@ class Component:
     logging) all of the parts that make up a Run, including: (1) the code of
     the object being run (i.e., the Component and its Entry Point), (2) the
     full DAG of other objects it depends on (i.e., DAG of other Components),
-    (3) the set of arguments (literally an ``ArgumentSet``) used during
-    initialization of the managed object and all objects it transitively
-    depends on, and (4) the arguments passed to the Entry Point being run.
+    (3) the set of arguments (literally an
```
nickjalbert, line 42: "literally an" -> "literally a"

web/registry/models.py
```diff
@@ -322,3 +323,50 @@ def display_string(self):
             f"Total Training Transitions: {self.training_step_count_metric}, "
             f"Mean Reward: {self.mean_reward_metric}"
         )
+
+    @staticmethod
+    def create_from_request_data(request_data: QueryDict):
+        pprint.pprint(request_data)
```
nickjalbert, line 329: Do we need this `pprint` here?

nickjalbert (approved):
Awesome!   Was able to run the demo script and publish my agent run locally (and also build the docs).  

A few comments, no blockers.  Feel free to merge!

agentos/agent_run.py
```diff
@@ -56,60 +58,84 @@ class AgentRun(Run):
...
-            trained. Defaults to "agent".
-        :param environment_name: The name of the environment component being
-            evaluated or trained. Defaults to "environment".
+        :param parent_run: Optionally, specify another Run that this run is
+            a sub-run of. Setting this will result in this AgentRun being
+            visually nested under the parent_run in the MLflow UI.
+        :param agent_identifier: Identifier of Agent component being evaluated
+            or trained.
+        :param environment_identifier: Identifier of Environment component
+            being evaluated or trained.
```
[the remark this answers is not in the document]
andyk: Done.

documentation/demos/index.rst
andyk: Done.

documentation/overview.rst
andyk: Done.

example_agents/acme_dqn/agent.py
andyk: Instead of making users import & use constants, which makes the API more cumbersome to users, I added two convenience factory class functions to AgentRun: `AgentRun.learn_run()` and `AgentRun.evaluate_run()`.

pcs/component.py
andyk: Done.

web/registry/models.py
andyk: Nope, was leftover debugging cruft.

PR #358, opened 2022-04-13
Fix dulwich checkout

### LINKED ISSUE #348: Our dulwich code can leave undesired files in the repo directory
The problem is that we initially clone into the default branch (e.g. master or main usually) when we clone a directory, but then we may checkout another branch (via a `porcelain.reset()`).  However this checkout/reset is buggy and/or unintuitive in that it doesn't clean up files that exist in master but do not exist in the branch you checked out.

See `pcs/repo.py` for the part of the code where we use porcelain.

This bug was mentioned in #342.

Here is a script to reproduce the problem:
```python
from pathlib import Path

from dulwich import porcelain
from dulwich.objectspec import parse_commit

local_repo_path = Path("dulwich_bug")
assert not local_repo_path.exists(), "Delete ./dulwich_bug dir"
local_repo_path.mkdir(parents=True)
github_url = "https://github.com/agentos-project/agentos.git"
porcelain.clone(source=github_url, target=str(local_repo_path), checkout=True)
repo = porcelain.open_repo(local_repo_path)
to_checkout = "07bc71358b4360092b58d78f9eee6dc939e90b10"
treeish = parse_commit(repo, to_checkout).sha().hexdigest()
porcelain.reset(repo=repo, mode="hard", treeish=treeish)

# The hash we checkout does not have a pcs/ folder so no files should exist.
# Confirm here: https://github.com/agentos-project/agentos/tree/07bc71358b4360092b58d78f9eee6dc939e90b10
# The problem is that we initially clone into master (which does have a pcs/
# folder), then we reset to a commit that doesn't have the pcs/ folder
# BUT dulwich doesn't do what you'd expect and clean up the non-existent dir
test_path = local_repo_path / "pcs" / "component.py"
assert not test_path.exists(), f"The path {test_path} should not exist!"
```

nickjalbert:
Fixes #348 

I did a little digging in dulwich and found that someone had seemingly fixed our checkout issue in a partially-unmerged pull request [here](https://github.com/jelmer/dulwich/pull/898).  I wrote a regression test, ported their `checkout()` code into our utils, and now things seem to be working correctly.

Regression demo:
```bash
# regression test without my fix; should fail
git checkout baae07da1b5cd2da6518d394313670f146b448dc
agentos clear-cache -y
pytest tests/test_repo.py::test_repo_checkout_bug

# regression test should pass with my fix applied
git checkout nj_dulwich_fix_2
agentos clear-cache -y
pytest tests/test_repo.py::test_repo_checkout_bug
```

### DIFF
--- a/pcs/repo.py
+++ b/pcs/repo.py
@@ -16,7 +16,7 @@ from pcs.exceptions import BadGitStateException, PythonComponentSystemException
 from pcs.identifiers import ComponentIdentifier, RepoIdentifier
 from pcs.registry import InMemoryRegistry, Registry
 from pcs.specs import NestedRepoSpec, RepoSpec, RepoSpecKeys, flatten_spec
-from pcs.utils import AOS_GLOBAL_REPOS_DIR, clear_cache_path
+from pcs.utils import AOS_GLOBAL_REPOS_DIR, clear_cache_path, dulwich_checkout
 
 logger = logging.getLogger(__name__)
 
@@ -361,7 +361,7 @@ class GitHubRepo(Repo):
         if treeish is None:
             treeish = parse_commit(repo, to_checkout).sha().hexdigest()
 
-        porcelain.reset(repo=repo, mode="hard", treeish=treeish)
+        dulwich_checkout(repo=repo, target=treeish)
         os.chdir(curr_dir)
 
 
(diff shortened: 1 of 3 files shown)


nickjalbert:
@andyk ready for you to take a look!

andyk (approved):
LGTM!

PR #360, opened 2022-04-14
Add codespell and pyupgrade to our code formatting

nickjalbert:
I was taking a look at the [gym repo](https://github.com/openai/gym/blob/master/.pre-commit-config.yaml) and noticed that they were using [codespell](https://github.com/codespell-project/codespell) and [pyupgrade](https://github.com/asottile/pyupgrade) in addition to the other static formatting and linting packages we already use (black, isort, and flake8).  

I love me a good formatter and/or linter, so I added the ones we were missing.


andyk (approved):
LGTM!

PR #369, opened 2022-04-19
Improve homepage view of runs

andyk:
Only show the terminal run in each trainining_run lineage DAG submitted per each environment


nickjalbert (approved):
LGTM!  Feel free to merge once it's green!

I suspect this'll be an early target for optimization (i.e. we won't want to pull in all runs and compute this on the fly every time index is loaded once we have a bit of data in the DB), but seems premature to worry about now!

andyk:
Yeah. I agree, this is definitely demo code! Can clean it up & optimize later :D

PR #375, opened 2022-04-21
[WIP] fix repo issues

nickjalbert:


nickjalbert:
This branch used the GitHub API to figure out things like "does this commit exist on the remote?".  Closing for now because we decided to use dulwich instead.  See #379

PR #379, opened 2022-04-26
Always use hashes for cached repos; dulwich refactor

nickjalbert:
This PR doesn't change any public functionality, just how we use dulwich to manage the repo cache.  Main changes:

*  I collected all the dulwich functionality scattered around `repo.py` and `utils.py` and dropped it into a `GitManager` class that `repo.py` now uses for lower-level git functions.
* We map all "versions" (sha1 hashes, branch names, or tag names) to a sha1 hash and then create a repo specifically for that hash in the cache.
* We also checkout a "default repo" for each GitHub project which isn't used at PCS runtime other than to map tag and branch names to sha1 hashes and to check what commits exist on the remote.
* We occasionally run a fetch on the default repo (once per process touching it and/or every 30 minutes in the case of long running processes) to get updated refs.

I'm going to rebase my papag work branch and make sure this addresses the repo issues we've been seeing.

### DIFF
--- a/pcs/git_manager.py
+++ b/pcs/git_manager.py
@@ -37,8 +37,7 @@ class GitManager:
 
         1. The file/directory is not in a git repo.
         2. The origin of this repo is not GitHub.
-        3. The current local branch and corresponding remote branch are not at
-           the same commit.
+        3. The current local commit is not in the remote repo.
         4. There are uncommitted changes locally
 
         If ''force'' is True, checks 2, 3, and 4 above are ignored.
@@ -224,7 +223,8 @@ class GitManager:
             with open(default_info_path, "w") as fout:
                 json.dump({PROC_KEY: os.getpid(), TIME_KEY: time.time()}, fout)
             print(f"GitManager: fetching {default_repo_path}...")
-            porcelain.fetch(porcelain.open_repo(default_repo_path))
+            with porcelain.open_repo_closing(default_repo_path) as repo:
+                porcelain.fetch(repo)
 
         if not default_info_path.exists():
             _do_fetch()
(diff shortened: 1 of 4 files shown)


pcs/git_manager.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def clone_repo(
+        self, org_name: str, project_name: str, version: str
+    ) -> Path:
+        """
+        Given a GitHub org, project name, and version (a sha1 commit hash,
+        branch name, or tag name), this will clone the repo at the version
+        into the PCS/AOS cache.
+        """
+        clone_destination = (
```
andyk, line 263: Do we want to resolve any branch or tag names to the githash (which it looks you're doing in `_checkout_version()` currently) and create a directory with that name in our local cache? Or is there a chicken and egg thing happening here where we need to checkout some version before we can use the repo to resolve the branch to a specific hash?

I guess this would be more of an optimization anyway, in the rare case where two different references point to the same underlying hash I think the current approach would just have two different directories that have the same contents.

andyk (approved):
Looks good! Left a few comments but nothing critical. Feel free to merge! Thanks for the fixes!

pcs/git_manager.py
```diff
@@ -0,0 +1,488 @@
...
+        """
+        PROC_KEY = "process_id"
+        TIME_KEY = "last_fetch_time"
+        default_info_path = Path(str(default_repo_path) + "-info.json")
+
+        def _do_fetch():
+            with open(default_info_path, "w") as fout:
+                json.dump({PROC_KEY: os.getpid(), TIME_KEY: time.time()}, fout)
+            print(f"GitManager: fetching {default_repo_path}...")
+            porcelain.fetch(porcelain.open_repo(default_repo_path))
```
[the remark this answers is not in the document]
nickjalbert: Ah, good call.  That seems better!

pcs/git_manager.py
nickjalbert: This is a good point, right now we're relying on the callers to resolve refs to hashes, but there's not reason not to do it right here.  Will update to not rely on the benevolence of the callers.

nickjalbert:
Alright, comments addressed, build is green, and I made sure `./documentation/demos/demo_i
lya_papag_from_cli.sh` is working locally.  Merging!

Thanks for the review!  Also, you might want to clear your AOS caches (`agentos clear-cache -y`).

PR #380, opened 2022-04-26
PAPAG and SB3 benchmark against the same environment component

nickjalbert:
This PR makes PAPAG and SB3 play against the same CartPole and Pong environments.  (Note: this branch is based on the branch from #381 so you can ignore the changes to `pcs/git_manager.py` since those changes are specifically up for review there).

You can run the demo script (you may need to clear your caches if you haven't since #379 was merged: `agentos clear-cache -y`):

```bash
# Run a clean webserver in one tab, and then in another tab in agentos root
 ./documentation/demos/demo_ilya_papag_from_cli.sh
```

Alternatively, you can run these agents on their own. 

PAPAG:

```bash
# Start fresh
cd example_agents/papag
rm -rf mlruns/

# CartPole learn and evaluate
agentos run agent --entry-point learn --arg-set-file ppo_cartpole_args.yaml
agentos run agent --entry-point evaluate --arg-set-file ppo_cartpole_args.yaml

# Pong learn and evaluate
agentos run agent --entry-point learn --arg-set-file ppo_pong_args.yaml
agentos run agent --entry-point evaluate --arg-set-file ppo_pong_args.yaml
```

SB3:
```bash
# Start fresh
cd example_agents/sb3_agent
rm -rf mlruns/

# CartPole learn and evaluate
agentos run sb3_agent --entry-point learn
agentos run sb3_agent --entry-point evaluate

# Pong learn and evaluate
agentos run sb3_agent --entry-point learn --arg-set-file ppo_pong_args.yaml
agentos run sb3_agent --entry-point evaluate --arg-set-file ppo_pong_args.yaml
```
Some notes:

* ~~I pinned the `gym[atari]==0.19.0` requirement because later `gym` versions moved the Atari ROMs out of the package and into subordinate packages.  This complicates selecting a particular class to make into an environment component, so I'm just sidestepping this for now by requiring an earlier `gym` version.~~  I pinned gym to 0.21.0 because rllib doesn't support the newer releases: https://github.com/ray-project/ray/issues/23062

* I had to copy over some chunks of PAPAG code in `example_agents/papag/agent.py` and then modify it so that it would play nice with PCS.  That code is fairly complicated and I had to jump through some extra hoops to make serialization work as PAPAG expected.


nickjalbert:
@andyk this is ready for review!

nickjalbert:
Uh oh, the Windows Gods have not smiled upon us today.  Looks like some issue with loading Pong; will debug.

nickjalbert:
@andyk Alright, this should be ready to go!

Here's the result on the local webserver after running `./documentation/demos/demo_ilya_papag_from_cli.sh`:

Screenshot 2022-04-29 at 17-56-58 AgentOS Leaderboard

example_agents/papag/agent.py
```diff
@@ -69,11 +91,12 @@ def evaluate(
         cuda=False,
     ):
         num_processes = int(num_processes)
+        env_class, _ = self._get_env_class_and_kwargs()
         with self.PAPAGRun.evaluate_run(
             outer_run=active_component_run(self),
             model_input_run=self.model_input_run,
-            agent_identifier=self.__component__.identifier,
-            environment_identifier=self.__component__.identifier,
+            agent_identifier=self.Algo.__component__.identifier,
```
andyk, line 98: We can discuss more if necessary, but in my PR against this PR (i.e. https://github.com/nickjalbert/agentos/pull/10) I revert this change to the agent_identifier because pointing at the algo component instead of the agent component makes it harder to differentiate the different "agents" that show up on the leaderboard. Specifically, when we publish a run of the sb3 agent using the PPO algo and a run of the PAPAG agent using the PPO algo, the table shows PPO in the "Agent" column for both runs.

tests/utils.py
```diff
@@ -1,50 +1,68 @@
...
+PAPAG_AGENT_DIR = EXAMPLE_AGENT_DIR / "papag"
+SB3_AGENT_DIR = EXAMPLE_AGENT_DIR / "sb3_agent"
+TEST_VENV_AGENT_DIR = ROOT_DIR / "tests" / "test_agents" / "venv_agent"
+
+TESTING_GITHUB_ACCOUNT = "agentos-project"
+TESTING_GITHUB_REPO = "agentos"
+TESTING_GITHUB_REPO_URL = (
+    f"https://github.com/{TESTING_GITHUB_ACCOUNT}/{TESTING_GITHUB_REPO}"
+)
+TESTING_BRANCH_NAME = "test_staging"
```
andyk, line 25: It's weird that the Github diff is shows that the first 25 lines of this file were deleted and added but the first 25 lines of the file that were added appear to be identical to the ones that were deleted. I wonder why the diff isn't only showing the updated lines starting at line 28.

Am I missing something (e.g., did white space change)?

andyk (changes requested):
@nickjalbert I hit some small issues when I tried to reproduce your results so created https://github.com/nickjalbert/agentos/pull/10 with some fixes. After that patch, I can run the script and see the same results in the web UI as you captured in your screenshot.

In my patch I also changed agent component that we pass to our PAPAGRun to be the `SB3PPOAgent`'s components instead of its `Algo`'s component. I did this so that we would end up with more descriptive component names for the two agents in the Leaderboard table, instead of having them both be called "PPO".

I think this is ready to go once we get my "PR PR" merged.

I'm also seeing some bugs in the webapp run view logic that I reported in #386, and I have an open PR built on top of https://github.com/nickjalbert/agentos/pull/10 that addresses them.

example_agents/papag/agent.py
nickjalbert: Yeah makes sense.  I think the trade-off here is that pointing the agent_identifier to the algo component is nice because then that component links directly to a PPO implementation in the papag or sb3 GitHub repos when you look at the component details in the web view.  

As it is, happy to accept the change for now and we can fiddle with this and/or the web views going forward.

tests/utils.py
nickjalbert: Yeah, for some reason this file had windows line terminators, so I converted it to unix terminators.  I assume this is causing the weirdness you're seeing!

PR #381, opened 2022-04-27
Git manager optimization

nickjalbert:
Small optimization to the `GitManager`.  Because we've already cloned the repo in the `_default` cache, we don't actually have to pull from GitHub to clone the repo at a particular sha1.  Instead, we just clone the local default repo and checkout at the hash.

### DIFF
--- a/pcs/git_manager.py
+++ b/pcs/git_manager.py
@@ -4,6 +4,7 @@ import re
 import time
 from json.decoder import JSONDecodeError
 from pathlib import Path
+from typing import Optional
 
 from dulwich import porcelain
 from dulwich.errors import NotGitRepository
@@ -25,6 +26,8 @@ class GitManager:
     """
 
     FULL_GIT_SHA1_RE = re.compile(r"\b[0-9a-f]{40}\b")
+    REPO_INFO_PROC_KEY = "process_id"
+    REPO_INFO_TIME_KEY = "last_fetch_time"
 
     def get_public_url_and_hash(
         self, full_path: Path, force: bool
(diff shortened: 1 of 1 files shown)


pcs/git_manager.py
```diff
@@ -204,39 +219,56 @@ def _get_default_repo_path(self, org_name: str, project_name: str) -> Path:
...
-            except JSONDecodeError:
-                _do_fetch()
-                return
-        sec_since_last_fetch = time.time() - info[TIME_KEY]
-        if info[PROC_KEY] != os.getpid() or sec_since_last_fetch > 1800:
+        if repo_info[self.REPO_INFO_PROC_KEY] != os.getpid():
+            _do_fetch()
+            return
+        sec_since_last_fetch = time.time() - repo_info[self.REPO_INFO_TIME_KEY]
+        if sec_since_last_fetch > 1800:
```
andyk, line 271: Can this be an `elif` so that we don't do the fetch twice in a row when both conditions are true?

andyk (approved):
This is neat! I like how you've implemented it :-D. Feel free to merge!

I am wondering how hard it will be for a dev to trigger the fetch w/ the force flag. Like should there a CLI command to allow a dev to pull changes in case they've pushed updates to the repo in the last 30min and want them fetched. Or does the dev have to open a Python REPL, import the git_manager and run `clone_repo()` with `force=True`, which seems like it could be slightly cumbersome?

andyk (commented):
Also, line 274 is a print statement ("GitManager: NOT fetching {default_repo_path}...") that we could probably delete.

pcs/git_manager.py
nickjalbert: I think this should be okay because we return immediately from `_maybe_fetch()` if we ever end up calling `_do_fetch()`  so no opportunity for a double fetch (it's hard to see in the diff view but clearer in the code, I think).

nickjalbert:
Ah, yeah: will remove that noisy log statement about NOT fetching

nickjalbert:
Removed that print statement, and I think we're safe from double fetching for now.  Build is green--merging!

nickjalbert:
Thanks for the review!

PR #387, opened 2022-05-01
Fix webregistry server run detail & make top corner logo link to homepage

### LINKED ISSUE #385: The logo on the WebRegistry webserver should go to home, not to the agentos docs
The de facto standard UX for clicking the logo in the top left of a website is to take you to the website homepage (the leaderboard view in our case), not some other website. We should change where our logo hyperlinks to.

### LINKED ISSUE #386: WebRegistry web server run detail view not correctly filtering runs shown in Run History
The webapp run detail view shows AgentRuns from runs that are not part of the current run's lineage, instead of correctly filtering down to just the runs that were part of the DAG that includes the currently displayed run.

andyk:
This is based on top of https://github.com/nickjalbert/agentos/pull/10 which itself is a PR on top of the branch in #380.

* Fixes #385
* Fixes #386
* Adds a new page that shows a prettier view of all runs in the registry in two tables: one for AgentRuns, and one for ComponentRuns. Also updates the "Runs" link in the main nav bar to point to this new view.

Screenshot of the new run list page:

Screen Shot 2022-05-02 at 16 49 47

### DIFF
--- a/web/leaderboard/views.py
+++ b/web/leaderboard/views.py
@@ -8,7 +8,7 @@ from registry.models import Component, ComponentDependency, Repo, Run
 
 
 def index(request):
-    run_obj_by_id, env_obj_by_id, _, terminals, _ = Run.agent_run_dags()
+    run_obj_by_id, env_obj_by_id, _, terminals, _, _ = Run.agent_run_dags()
     env_dict = defaultdict(list)
     for env_id, run_list in terminals.items():
         runs = sorted(
--- a/web/leaderboard/views.py
+++ b/web/leaderboard/views.py
@@ -1,5 +1,4 @@
 from collections import defaultdict
-from datetime import datetime
 
 from django.conf import settings
 from django.http import HttpResponseBadRequest, HttpResponseRedirect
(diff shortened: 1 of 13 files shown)


nickjalbert:
@andyk could you close, rebase on master, and re-open when you get a chance?  GitHub is doing that thing where the diff is outdated and not updating :(

andyk:
Okay @nickjalbert, I didn't close/rebase/reopen, but I did merge master into the branch this PR is based and now the diff view seems to be showing me the correct thing.

Is it working for you too?

web/leaderboard/views.py
```diff
@@ -18,17 +19,43 @@ def index(request):
...
 
+def run_list(request):
+    agent_runs = Run.objects.filter(
+        data__tags__contains={"pcs.is_agent_run": "True"}
+    )
+    component_runs = Run.objects.filter(
+        data__tags__contains={"pcs.is_component_run": "True"}
+    )
+    component_runs_with_starttimes = {}
+    for run in component_runs:
```
nickjalbert, line 34: I think a more canonical way to achieve this is to put something like the following on the Run model:

```python
@property
def start_time(self):
    start_time_str = self.info["start_time"]
    start_time = datetime.fromtimestamp(int(start_time_str) / 1000)
    return start_time.strftime("%d/%m/%y %H:%m")
```

This, however, can end up being somewhat confusing because you can use `run.start_time` in templates, but it's not actually a column in the DB, and so the ORM will not be happy if you try to filter or sort using `start_time` (even though it's indistinguishable from other columns when you're using it on a `Run` object).

I think either way is ok for now, but we should probably move it onto the `Run` object if we end up wanting to use this in more places.

nickjalbert (approved):
Nice, web is coming along. :)

This PR LGTM!  Feel free to address or ignore my comment.

Some thoughts not necessarily for this PR.  I ran the following commands on your branch to test locally:

```bash
cd example_agents/sb3_agent
rm -rf mlruns/
agentos freeze sb3_agent --output-file frozen_sb3.yaml
agentos run sb3_agent --registry-file frozen_sb3.yaml --entry-point learn
agentos run sb3_agent --registry-file frozen_sb3.yaml --entry-point learn
agentos run sb3_agent --registry-file frozen_sb3.yaml --entry-point learn
agentos run sb3_agent --registry-file frozen_sb3.yaml --entry-point evaluate
USE_LOCAL_SERVER=True agentos publish-run <EVAL AGENT RUN ID>
```
and ended up getting the following locally on the run detail page:

Screenshot 2022-05-12 at 11-42-52 Run b3a477c6f1c14984b4714acf44d552e3

It might make sense to:

* Sort the learning runs by the time completed (or started?) descending.  Here's where actually having the timestamp in the DB might be useful (although, I think Postgres lets us reach into JSON fields now, so you may be able to do this without creating a new column).  Alternatively, you could sort in the view function or in the template itself.
* Only show Runs with `run_type=train` in the learning history (instead of both `train` and `learn`)

Feel free to punt on this for now if you'd prefer.  

Once we merge, I can push an update to Heroku.

nickjalbert:
btw merging master and pushing fixed the diff for me.  Thanks!

PR #391, opened 2022-05-28
Update CartpoleEnv version to match gym==0.21.0

nickjalbert:
Minor fix, This updates the `CartPoleEnv` we pull from the gym repo to match the version of the gym (==0.21.0) that we install in the environment from the PAPAG/SB3 requirements.txt.   We were using the `CartPoleEnv` from 0.19.0.  

See the listing of gym tags here: https://github.com/openai/gym/tags

### DIFF
--- a/example_agents/papag/components.yaml
+++ b/example_agents/papag/components.yaml
@@ -25,7 +25,7 @@ components:
         dependencies:
             PAPAGRun: PAPAGRun
             AtariEnv: AtariEnv==db3728264f382402120913d76c4fa0dc320ef59f
-            CartPoleEnv: CartPoleEnv==4ede9280f9c477f1ca09929d10cdc1e1ba1129f1
+            CartPoleEnv: CartPoleEnv==c755d5c35a25ab118746e2ba885894ff66fb8c43
             A2C_ACKTR: A2C_ACKTR==41332b78dfb50321c29bade65f9d244387f68a60
             PPO: PPO==41332b78dfb50321c29bade65f9d244387f68a60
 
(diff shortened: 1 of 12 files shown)


example_agents/sb3_agent/components.yaml
```diff
@@ -24,7 +24,7 @@ components:
         requirements_path: ./requirements.txt
         dependencies:
             AtariEnv: AtariEnv==db3728264f382402120913d76c4fa0dc320ef59f
-            CartPoleEnv: CartPoleEnv==4ede9280f9c477f1ca09929d10cdc1e1ba1129f1
+            CartPoleEnv: CartPoleEnv==c755d5c35a25ab118746e2ba885894ff66fb8c43:
```
andyk, line 27: Same here. Did you intend for there to be a colon at the end of this line? Seems like these colons might be breaking the sb3 tests (which seem to be failing when this file is parsed).

example_agents/papag/components.yaml
```diff
@@ -41,7 +41,7 @@ components:
         class_name: AtariEnv
         instantiate: False
 
-    CartPoleEnv==4ede9280f9c477f1ca09929d10cdc1e1ba1129f1:
+    CartPoleEnv==c755d5c35a25ab118746e2ba885894ff66fb8c43:
```
andyk, line 44: Not sure why this line would have a colon at the end?

andyk (changes requested):

example_agents/papag/components.yaml
nickjalbert: This one needs a colon, right?  Because it's the key with value {repo: ..., file_path: ..., ...}

example_agents/sb3_agent/components.yaml
nickjalbert: This, however, is a typo!

example_agents/papag/components.yaml
andyk: Ah right. My bad (sorry reviewing on phone)

nickjalbert:
Okay, got the tests to green by pinning protobuf to the version released in April 2022.  Not sure why the newest protobuf isn't playing nice with mlflow on mac, but something we can dig into later if we're feeling ambitious.  Merging!

Thanks for the review!

PR #395, opened 2022-06-08
fix more tests - follow-on to spec revamp

andyk:
Fixes more tests. Get build back to green and uncomment out some of the tests that #336 left broken


andyk:
@nickjalbert I guess we might want to get the tests runtime back to a few min (from a few hours 🤪) before we merge

nickjalbert:
Can't argue with green, though!

Issue #303, opened 2022-03-05
Add PyPIRepo


andyk:
This issue depends on #278.

A PyPIRepo will represent a PyPI project, which is a container of versioned packages.

This will let a Module Component (after #278) explicitly document a module dependency as an entry in it's `dependencies` dictionary.

PR #397, opened 2022-06-09
Publish an AgentRun to a WebRegistry

nickjalbert:
This PR allows us to publish an AgentRun to the WebRegistry.  How I ended up here:

* I was able to publish the Output of the SB3 agent to the WebRegistry.  However, we store all the goodies (like environment and agent IDs) on the child AgentRun
* I updated `agentos run` to print more info about the Output and its children
* I tried to publish a child AgentRun, but kept running into issues with inconsistencies in the generation of AOS/PCS objects from existing MLflow runs (i.e. sometimes we allowed passing of `existing_run_id` in the object `__init__()`, sometimes not)
* Now each run type (MLflowRun, Output, AgentRun) essentially implements two methods: one to initialize a new run (`__init__()`) and another to initialize the object from an existing MLflow run in `mlruns/` (`from_existing_mlflow_run()`).  They share some creation and validation logic.

In this PR, the following script works and pushes the AgentRun as well as the Output and its children recursively:

```bash
# Run a web server in a tab, empty the DB
# In a new tab
cd example_agents/sb3_agent
rm -rf mlruns/
agentos run sb3_agent
USE_LOCAL_SERVER=True agentos publish <AgentRun ID from previous command>
```

Questions and TODOs:
* Maybe we should instead just have an `__init__()` that handles both cases (new MLflow run, MLflow run from disk) and have a switch based on the scenario we're in
* The AgentRun doesn't have a spec:link to its parent Output (you can trace it via the tags, but we probably want this stored in the DAG itself)


andyk:
Would it simplify things if we had the mlflow_run be a member of an AgentRun rather than use inheritance?

agentos/agent_output.py
```diff
@@ -25,6 +25,30 @@
 
 RunStats = namedtuple("RunStats", _RUN_STATS_MEMBERS)
 
+SPEC_ATTRS = []
```
andyk, line 28: I know you mentioned this in your comment, but is there any reason these couldn't be `@classmethod`s?

andyk (approved):
Overall, think this looks good and you should free to merge so we can focus on getting the rest of the webapp logic ported over.

It does feel a little weird to have these global functions living outside of the classes, but as we've discussed a bunch of times, our integration with MLflow needs another revamp anyway. And I don't think it is more important than some of our other priorities, such as:
1. getting the webapp logic ported and getting back to focusing on our RL benchmark/comparison demos
2. prototyping our low/no-code solution for auto-generating registries (including work on #303)

Thus, at a high level I think we should merge this (or something that is not too much more work to refactor this) and push forward!

andyk:
I have been starting to think that maybe we should think of MLflow less as an integral part of PCS and more as a pluggable storage backend + visualization tool which PCS can be configured to write/read its registries to/from.

If we thought if it more like that, then maybe we could basically support `Component.to_mlflow()` and `.from_mlflow()` methods. Or another option would be for the MLflowRun component to be a shallow pointer such as:
```yaml
my_mlflow_run_spec:
    type: MLflowRun
    mlflow_tracker_url: http://localhost:5000
    mlflow_run_id: 28c7fe4c989d7f3f7fe489d7fe4374a8
```

agentos/agent_output.py
nickjalbert: Yep, I was able to move these functions into the classes.

agentos/agent_output.py
```diff
@@ -25,6 +25,30 @@
```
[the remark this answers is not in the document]
nickjalbert: Fixed.

nickjalbert:
Yeah, the pluggable MLflow backend sounds like a promising way to go!  Moved the global functions into the classes and renamed `agent_output.py` -> `agent_run.py`.  Will merge once tests are green.

Thanks for the review!

andyk:
🤗

PR #405, opened 2022-06-18
Update example agent to use AgentRun

nickjalbert:
This updates the default agent (created by `agentos init`) to use `AgentRun` so that it's compatible with the web frontend.  This is part of a larger push to automatically populate the web with data for testing and demo purposes.

This also updates `example_agent/random` to match the default agent.

Demo:

```bash
# Start the local web server in a separate tab
# Clear the local DB by visiting http://localhost:8000/empty_database in your browser
mkdir demo
cd demo
agentos init .
agentos run agent
USE_LOCAL_SERVER=True agentos publish <agent run ID printed out after previous previous command>
# Now visit http://localhost:8000 and see the result of your run
```

The leaderboard will show something like the following after running the demo script:

Screenshot 2022-06-18 at 16-52-56 AgentOS Leaderboard


andyk (approved):
Looks great! I checked out this branch and tried out `init`ing, running, and publishing a new template agent, as well as running & publishing the random agent. All worked well. I'm going to merge and then make a PR that cleans up the output text a little bit.

PR #407, opened 2022-06-20
Silence dulwich spam in github tests

nickjalbert:
This PR silences the dulwich spam (e.g. "Counting objects 1/3000") in the github test logs.  We still have a lot of noise, but now it's from PCS/AOS proper and we clean that up as desired.

[Example logs with this change](https://github.com/agentos-project/agentos/runs/6965505960?check_suite_focus=true)


andyk:
Thanks for doing this

PR #413, opened 2022-06-22
Port web tests to new models

nickjalbert:
This PR ports the web tests to the new models.  This PR:

* Contains two main integration tests: `test_run_and_publish` and `test_web_registry_integration`.
    * `test_run_and_publish` creates the default agent (`agentos init`), runs it, and then uploads the run to the registry.  It also saves the resulting database as a fixture in  `web/registry/tests/fixtures/` so that the other tests can quickly populate the database with realistic data.
    * `test_web_registry_integration` tests that we can create a `WebRegistry` backed by a web server and that we can put and get specs via that `WebRegistry`.
* Adds a `DEFAULT_REGISTRY_URL` environment variable for specifying backing web servers other than localhost and heroku.
* Adds tests for the view we're using in the `leaderboard` app.
* Resurrects the `Component.artifact_tarball` file field and related endpoints to get/put artifacts related to the a `Component`.  We don't actually use this yet when publishing (and maybe we'll take a different approach to sharing artifacts?).


web/registry/tests/test_integration.py
```diff
@@ -1,203 +1,39 @@
...
+    def test_web_registry_integration(self):
+        load_or_create_fixture(self.live_server_url)
+        # Check WebRegistry mirrors server
+        registry = WebRegistry(f"{self.live_server_url}/api/v1")
+        self.assertEqual(Component.objects.count(), 20)
+        for component in Component.objects.all():
+            registry.get_spec(component.identifier, flatten=False)
+            registry.get_spec(component.identifier, flatten=True)
+        # Add new spec to WebRegistry, make sure it makes it to the server
+        gh_url = "https://github.com/chipsalliance/firrtl-spec"
```
andyk, line 34: Out of curiosity, any reason you picked this repo for tests?

andyk (approved):
Thanks for fixing up these tests!

PR #420, opened 2022-07-07
Upload artifacts when publishing AgentRuns

nickjalbert:
This PR gets artifact upload working again.  

I think this is the shortest path to getting us back to sharing models; this is basically just porting forward the upload system to the new spec revamp.  

Longer term, we should think more about artifacts being explicitly represented in the DAG (maybe something like a file-in and file-out component?) rather than doing this via a "side-channel" as we do here.

Demo:

```bash
# run the web server in another tab; clear the DB
cd example_agents/sb3_agent
rm -rf mlruns/
agentos run sb3_agent --function-name learn
agentos run sb3_agent --function-name evaluate
USE_LOCAL_SERVER=True agentos publish <id of agent run printed after previous command>
```

If you follow the demo, all the AgentRuns should have tarballs associated with them that contain the model that was trained/evaluated.  For example:

Screenshot 2022-07-08 at 16-28-16 Select component to change Django site admin

### DIFF
--- a/agentos/agent_run.py
+++ b/agentos/agent_run.py
@@ -1,12 +1,19 @@
+import json
+import shutil
 import statistics
+import tarfile
+import tempfile
 from collections import namedtuple
+from pathlib import Path
 from typing import Optional
 
+import requests
 from mlflow.entities import RunStatus
 from mlflow.utils.mlflow_tags import MLFLOW_PARENT_RUN_ID, MLFLOW_RUN_NAME
 
 from pcs.mlflow_run import MLflowRun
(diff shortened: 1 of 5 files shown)


nickjalbert:
@andyk this is ready for you to take a look at when you have time!

andyk:
This looks great! Thanks for getting this working again. I'm in favor of having this in place as a demo solution and potentially re-thinking in the future.

I tried this out per your instructions on my Macbook Pro M1 and it worked great! Merging.

andyk (approved):
LGTM!

PR #421, opened 2022-07-17
Adds VirtualEnv Component

andyk:
* Refactors VirtualEnv to be a Component
* Adds VirtualEnvModule Component
* Adds RelativePath Component and pcs.Path abstract base class. Makes Repo inherit from pcs.Path
* Renames Module -> FileModule
* Removes `requirement_path` attribute from Module and adds optional `virtual_env` attribute.
* Renames GitHubRepo -> GitRepo
* Moves version attribute from Module Component to GitRepo


andyk:
@nickjalbert This is ready for a review!

documentation/repl_quickstart.rst
```diff
@@ -20,12 +20,12 @@ In the example below we first define a very basic class called ``Adder``.
 Then we use PCS to track the execution of a method on this class and share the
```
nickjalbert, line 20: Looks like maybe there's a kruft file in this dir?  `documentation/demos/temp.py`

example_agents/acme_r2d2/components.yaml
```diff
@@ -18,10 +25,10 @@ specs:
             type: Class
             name: AcmeR2D2Agent
             module:
-                type: Module
+                type: FileModule
                 repo: spec:local_dir
                 file_path: ./agent.py
-                requirements_path: ./requirements.txt
+                virtual_env: spec:virtual_env
```
nickjalbert, line 31: Ah k, so I see the virtual_env pointer here.  Cool!

pcs/instance_manager.py
```diff
@@ -27,20 +26,24 @@ def __init__(
         self.register_attributes(["instance_of", "argument_set"])
         self._instance = None
 
-    def get_object(self):
+    def get_new_object(self):
         if self._instance:
             return self._instance
```
nickjalbert, line 31: Does it make sense to return the cached `_instance` in `get_new_object()`?  From the name of the method, I'd guess it'd give me a new object even if there's a cached instance.

pcs/module_manager.py
```diff
@@ -128,57 +63,83 @@ def get_object(self):
...
+class FileModule(Module):
+    def __init__(
+        self,
+        repo: Repo,
+        file_path: str,
+        imports: Dict[str, "Module"] = None,
+        virtual_env: VirtualEnv = None,
+    ):
+        """
+        :param repo: Repo where this Module's source file can be found. The
```
nickjalbert, line 79: High-level thought: it occurs to me that the `repo` terminology can be a bit misleading (esp. when the repo is just a local directory on the filesystem).  I wonder if there's a better word for this...

pcs/utils.py
```diff
@@ -281,3 +291,38 @@ def copy_find_and_replace_leaves(
...
             transform_leaf(data_struct, copy_struct, leaf_list, lambda x: x)
     return match_found, copy_struct
+
+
+def pad_list_if_necessary(list_in, index):
+    while index >= len(list_in):  # List too short
+        list_in.append(None)
+
+
+def pipe_and_check_popen(args: List[str], **kwargs):
```
nickjalbert, line 301: Nice!  Looks like you're using this to call out to pip.  Seems like a smarter (or at least more customizable) version of `subprocess.run`.  What prompted you to go about it this way?

nickjalbert (approved):
Looks great to me!  A few questions/comments but nothing big from my end during this pass; feel free to merge whenever and drop any follow-up fixes in a new PR.

~~I was having difficulty running SB3 directly from command-line~~.  Never mind--I had to clear my cache and remove `mlruns/` but now it's working like a charm.

documentation/repl_quickstart.rst
andyk: Good catch. Removed.

pcs/module_manager.py
andyk: Yeah. Maybe FileSource?

pcs/utils.py
andyk: I found myself tweaking the flags we pass to Popen so that we fail fast when our sub processes fail (e.g., pip install into a virtualenv) and the stdout/stderr show up to make debugging easier. And I figured DRYing out the various places where we use subprocesses (to run the `virtualenv` command vs to run the `pip` command) would be good hygiene.

